PR #110, opened 2026-06-17
fix: avoid out-of-bounds array range when a shift cuts through an array

jafin:
## Problem

Follow-up to #109. The array/dynamic-array shift handling added in #109 relocates an array formula's stored spill range when a row/column insert or delete moves the array's cells. When a **delete overlaps the array** (deleting part of it — which Excel forbids interactively but the XLibur API allows), the relocation subtracts the deleted count from the array's top edge and can drive the coordinate below 1.

`XLSheetPoint`'s constructor does not validate bounds (`(uint)(row - 1)`), so a non-positive coordinate silently overflows, corrupting the stored range.

Reproduction:

```csharp
ws.Range("A2:A4").FormulaArrayA1 = "TRANSPOSE({1,2,3})";
ws.Rows(1, 2).Delete();   // deletes through the array
// before fix: array range becomes A0:A2 — invalid row 0; on save the master
//             cell (A0) doesn't exist, so the <f t="array"> element is dropped.
```

## Fix

`ShiftArrayRangeRows`/`ShiftArrayRangeColumns` now relocate the range only when the array lies **entirely within the region the shift actually moves**:
- inserts (positive shift): the array must start at/after the insertion edge;
- deletes (negative shift): the array must start strictly **after** the deleted band (`lastRow + 1` / `lastColumn + 1`).

This guarantees the shifted coordinates stay `>= 1` and avoids relocating an array the operation cuts through. Cut-through cases leave the range unchanged (valid coordinates, saveable workbook) rather than corrupting it.

## Tests

- `DeletingRowsThroughArrayDoesNotCorruptRange`
- `DeletingColumnsThroughArrayDoesNotCorruptRange`

Both assert the surviving array range stays in-bounds and the workbook still saves. Full suite: **6052 passed, 0 failed**.

### DIFF
--- a/XLibur.Tests/Excel/CalcEngine/ArrayFormulaTests.cs
+++ b/XLibur.Tests/Excel/CalcEngine/ArrayFormulaTests.cs
@@ -337,4 +337,52 @@ public class ArrayFormulaTests
         var metadataXml = metadataReader.ReadToEnd();
         Assert.That(metadataXml, Does.Contain("fDynamic=\"1\""), "Dynamic-array metadata missing");
     }
+
+    [Test]
+    public void DeletingRowsThroughArrayDoesNotCorruptRange()
+    {
+        // Deleting rows that overlap an array used to push the stored range past row 1, producing
+        // an out-of-bounds coordinate (e.g. A0:A2) via the unchecked XLSheetPoint constructor.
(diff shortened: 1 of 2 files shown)


jafin:
@coderabbitai review

PR #142, opened 2026-07-12
fix: shrink named range when deleting its top row/left column

jafin:
## Summary

Deleting the **first row of a named range** shifted both endpoints upward instead of removing the deleted row and shifting the survivors up. A named range `A3:A4` with row 3 deleted became `A2:A3` — expanding the range to include a row (`A2`) that was never part of it — where Excel produces `A3:A3`.

```
Before: $A$3:$A$4
After (was):      $A$2:$A$3
After (Excel/fix): $A$3:$A$3
```

## Root cause

In `XLCellFormulaShifter`, when the deleted region starts at or before the range's first row, the code took the "range is entirely below the deleted region" branch and shifted **both** endpoints by the shift amount. That is only correct when the deletion is fully *above* the range. When the deletion overlaps the range's **top boundary** while rows below it survive, Excel clamps the top edge to the deletion start and shifts only the bottom edge up.

## Changes

- `XLCellFormulaShifter.AppendShiftedRowMatch` — new `IsTopBoundaryDeletion` guard routes the overlapping-top case to `AppendClampedTopRowShift` (leading edge → deletion start, trailing edge → shifted).
- Symmetric fix for the **column** path (`AppendShiftedColumnMatch` / `IsLeftBoundaryDeletion` / `AppendClampedLeftColumnShift`): deleting the left column of `C1:D1` now yields `C1:C1` instead of `B1:C1`.
- Both guards require a negative shift, so **row/column insertion behavior is unchanged**.

## Behavior

| Operation | Named range | Before | After |
|---|---|---|---|
| Delete top row | `A3:A4` | `A2:A3` | `A3:A3` |
| Delete rows 2:3 | `A3:A5` | `A1:A3` | `A2:A3` |
| Delete middle row 4 | `A3:A5` | `A3:A4` | unchanged |
| Delete row above (row 1) | `A3:A4` | `A2:A3` | unchanged |
| Delete left column | `C1:D1` | `B1:C1` | `C1:C1` |

## Testing

- New `NamedRangeDeletionTests` (7 cases): top-row shrink, multi-row top overlap, middle-row regression guard, row-above regression guard, multi-column block, left-column shrink, column-left regression guard.
- Full suite: **6095 passed, 0 failed**.

Refs [ClosedXML/ClosedXML#2866](https://github.com/ClosedXML/ClosedXML/issues/2866)

### DIFF
--- a/XLibur.Tests/Excel/NamedRanges/NamedRangeDeletionTests.cs
+++ b/XLibur.Tests/Excel/NamedRanges/NamedRangeDeletionTests.cs
@@ -1,0 +1,141 @@
+using System.Linq;
+using XLibur.Excel;
+using NUnit.Framework;
+
+namespace XLibur.Tests.Excel.NamedRanges;
+
+[TestFixture]
+public class NamedRangeDeletionTests
+{
+    private static string RefersTo(XLWorkbook wb, string name)
+    {
+        return wb.DefinedNames.First(dn => dn.Name == name).RefersTo;
+    }
+
+    /// <summary>
+    /// Regression for issue #2866. Deleting the top row of a named range must remove that row and shift the
+    /// survivors up (A3:A4 -> A3:A3), matching Excel. Previously ClosedXML shifted both endpoints upward,
+    /// expanding the range to A2:A3 and including a row that was never part of it.
+    /// </summary>
+    [Test]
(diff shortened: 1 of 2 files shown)


jafin:
@coderabbitai review

PR #186, opened 2026-07-25
docs: add feature guides for worksheets, tables, formulas and more

jafin:
## What

Adds 14 topic pages to the Docusaurus site, grouped into five sidebar categories:

| Category | Pages |
|---|---|
| Working with workbooks | `worksheets`, `cells-and-ranges`, `importing-exporting` |
| Formatting | `styling`, `theming`, `conditional-formatting`, `fonts` |
| Data features | `tables`, `autofilter`, `pivot-tables`, `data-validation` |
| Calculation | `formulas`, `functions` |
| Output | `page-setup` |

Each page opens with an overview of the topic, then works through short task-shaped examples, and ends with a worked end-to-end sample.

Also trims `introduction.md`: the "Should I use this?", "User Guide", "Building, Testing and Benchmarks", "Developer guidelines" and "Credits" sections duplicated the README and are better served by the new guides and the repository itself.

## Why

The site previously had two pages (Introduction, Getting Started) and pointed users at the ClosedXML documentation for everything else. That works for the API surface we inherited unchanged, but not for the parts where XLibur differs — most obviously the swappable font engine — and it gives no landing place for the day-to-day "how do I do X" questions.

## Accuracy

Every example was written against the XLibur source (`XLibur/Excel/**` interfaces and the `XLibur.Examples` project) rather than carried over from ClosedXML docs, so signatures, enum members and return types match this fork. Three behaviours that are easy to get wrong are called out explicitly rather than glossed over:

- **Workbook theme edits only persist for workbooks created from scratch.** `XLWorkbook_Save` writes a theme part only when the package does not already have one, so mutating `Theme` on a loaded file is silently a no-op. Documented as a warning on the Theming page, with the workaround.
- **Data table (what-if) formulas round-trip but cannot be created.** `XLCellFormula.DataTable1D`/`2D` are `internal`. The Formulas page says so and shows the template-based alternative.
- **Dynamic array functions need `SetDynamicFormulaA1`, not `FormulaA1`** — otherwise the formula is written as `=@FILTER(...)` and will not spill.

The supported-function list on the Functions page is derived from the `ce.RegisterFunction` call sites in `Excel/CalcEngine/Functions`, categorised by source file. `BAHTTEXT`, `JIS` and `PHONETIC` are excluded because their registrations are commented out in `Text.cs`.

A couple of other sharp edges are documented where they belong: AutoFilter's `AddFilter` compares *formatted strings* using the current culture (so it is locale-sensitive for numbers), filter column indexes are relative to the filter range rather than the sheet, and worksheet names are not validated by the library.

## Verification

- `pnpm build` — passes. `onBrokenLinks` is `'throw'`, so every cross-link between pages resolves.
- `pnpm typecheck` — clean.
- All 14 pages confirmed present in `build/`.


jafin:
Pushed a second commit covering the areas the first pass missed. Seven more pages, plus a new **Visuals** sidebar category:

| Page | Covers |
|---|---|
| `charts` | 70+ chart types, series references, two-cell anchoring, combo charts, extended types |
| `sparklines` | In-cell line / column / win-loss groups, markers, axis scaling, date axis |
| `images` | Embedding pictures, the three anchor styles, placement, scaling, grouping, extraction |
| `comments-and-hyperlinks` | Cell notes with rich text and styling; internal and external links |
| `defined-names` | Named ranges, workbook vs worksheet scope, named constants, broken-reference auditing |
| `grouping-and-outlines` | Collapsible row/column outline levels, nesting, summary location |
| `workbook-settings` | Document and custom properties, protection, calculation mode, `LoadOptions` / `SaveOptions` |

**Charts are the notable gap.** ClosedXML 0.105 could not create charts at all, so there is no upstream documentation to fall back on — this is new API with no prior coverage anywhere. The page works through the full `XLChartType` catalogue by family, and documents the data-shape requirements that differ per family: stock charts want the price columns in a fixed order, `Treemap`/`Sunburst` take a multi-column category reference for the hierarchy, `BoxWhisker` takes raw observations with a grouping column. It also states plainly what the API does *not* expose (axis titles, legend placement, data labels, series colours) and shows the template-workbook approach for when that control is needed.

Other sharp edges now written down:

- **Chart anchors are 0-based** — unlike the 1-based cell addressing used everywhere else in XLibur. `Position.Column` is written straight to `xdr:colId`.
- **Picture types are in `XLibur.Excel.Drawings`**, so they need a second `using` beyond `XLibur.Excel`.
- **Workbook protection defaults to `Algorithm.SimpleHash`**, not SHA512 — `SHA512` has to be passed explicitly.
- **Renaming a defined name does not rewrite formulas** that use it; `SetRefersTo` is the safe operation.
- **`Outline.SummaryVLocation`** must be `Top` when totals sit above their detail, or the collapse button appears on the wrong row.
- **Comments are legacy notes**, not Excel 2016 threaded comments — XLibur does not create the latter.

Also cross-linked the new pages from the existing guides, and switched the comment snippet in `cells-and-ranges` from `CreateComment()` to `GetComment()`: the snippet reads as "add a comment", but `CreateComment()` replaces any comment already present.

`pnpm build` passes (`onBrokenLinks: 'throw'`, no anchor warnings) and `pnpm typecheck` is clean. 21 doc pages total.

PR #225, opened 2026-07-26
fix: preserve automatic/absent colour on rich-text runs (#219)

### LINKED ISSUE #219: Rich-text runs with automatic/absent colour serialized as explicit rgb=FF000000 on round-trip
### XLibur Version

0.106.0 (via `XLibur.Bundle` 0.106.0)

### Target Framework

.NET 10

### Platform / OS

Windows

### Description

On a plain load → `SaveAs()` round-trip, rich-text runs whose formatting contains **no explicit `<color>` element** are written back with an explicit `<color rgb="FF000000"/>`. The library cannot distinguish "colour was never set" (which Excel renders using the theme/automatic colour, and which conditional formatting and theme colour changes can override) from "colour was explicitly set to black", so it materialises black on every colour-less run. This silently alters the original formatting.

There is a related second symptom: a shared string that is **plain text carrying a phonetic guide** (`<t>` + `<rPh>` + `<phoneticPr>`, i.e. no rich-text `<r>` runs at all — common in Japanese workbooks) is promoted into a synthetic rich-text run on save, and that run then inherits the same `rgb="FF000000"` injection.

This is the same behaviour reported upstream in ClosedXML — [ClosedXML#2872](https://github.com/ClosedXML/ClosedXML/issues/2872) — and I've confirmed XLibur 0.106.0 reproduces it byte-for-byte identically to ClosedXML 0.105.0 (same three injected `rgb="FF000000"` elements, same `vertAlign="baseline"` / `family="2"` additions, same phonetic promotion). Filing here so XLibur can track/fix it independently of upstream.

### Reproducible Test Case

The bug is about *reading* a file whose runs have no colour, so the input can't be produced through the `XLWorkbook` API (the writer would normalise it before we ever save). The test therefore builds a minimal, spec-valid `.xlsx` in memory (no external file needed), round-trips it, and asserts on the resulting `sharedStrings.xml`.

```csharp
using System.IO.Compression;
using System.Text;
using System.Text.RegularExpressions;
using NUnit.Framework;
using XLibur.Excel;

[TestFixture]
public class RichTextColorRoundTripTests
{
    // A minimal workbook whose sharedStrings.xml contains:
    //   si[0] - rich-text runs whose rPr has NO <color> element
    //   si[1] - plain text + phonetic guide (rPh/phoneticPr), i.e. NOT rich text
    // Neither carries any colour.
    private const string SharedStrings =
        """
        <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
        <sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="2" uniqueCount="2">
          <si>
            <r><rPr><b/><sz val="11"/><rFont val="Calibri"/></rPr><t>Bold</t></r>
            <r><t xml:space="preserve"> plain</t></r>
          </si>
          <si>
            <t>&#28450;&#23383;</t>
            <rPh sb="0" eb="2"><t>&#12363;&#12435;&#12376;</t></rPh>
            <phoneticPr fontId="0" type="Hiragana"/>
          </si>
        </sst>
        """;

    [Test]
    public void ColorlessRuns_AreNotSerializedAsExplicitBlack()
    {
        var input = BuildWorkbook();

        using var wb = new XLWorkbook(new MemoryStream(input));
        using var outMs = new MemoryStream();
        wb.SaveAs(outMs);

        var savedSharedStrings = ReadPart(outMs.ToArray(), "xl/sharedStrings.xml");

        // The source had no <color> anywhere; the round-trip must not invent black.
        var injectedBlack = Regex.Matches(savedSharedStrings, "rgb=\"FF000000\"", RegexOptions.IgnoreCase).Count;
        Assert.That(injectedBlack, Is.Zero,
            $"No explicit black colour was present in the source, but the save injected {injectedBlack}.\n\n{savedSharedStrings}");

        // And the plain-text-with-phonetics string must not be promoted to a coloured run.
        Assert.That(savedSharedStrings, Does.Contain("<rPh"),
            "Phonetic run (rPh) should be preserved.");
    }

    private static string ReadPart(byte[] xlsx, string partPath)
    {
        using var zip = new ZipArchive(new MemoryStream(xlsx), ZipArchiveMode.Read);
        var entry = zip.GetEntry(partPath) ?? throw new AssertionException($"Missing part: {partPath}");
        using var r = new StreamReader(entry.Open());
        return r.ReadToEnd();
    }

    private static byte[] BuildWorkbook()
    {
        var parts = new (string Path, string Content)[]
        {
            ("[Content_Types].xml",
                """
                <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
                <Types xmlns="http://schemas.openxmlformats.org/package/2006/content-types">
                  <Default Extension="rels" ContentType="application/vnd.openxmlformats-package.relationships+xml"/>
                  <Default Extension="xml" ContentType="application/xml"/>
                  <Override PartName="/xl/workbook.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet.main+xml"/>
                  <Override PartName="/xl/worksheets/sheet1.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.worksheet+xml"/>
                  <Override PartName="/xl/sharedStrings.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.sharedStrings+xml"/>
                  <Override PartName="/xl/styles.xml" ContentType="application/vnd.openxmlformats-officedocument.spreadsheetml.styles+xml"/>
                </Types>
                """),
            ("_rels/.rels",
                """
                <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
                <Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
                  <Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/officeDocument" Target="xl/workbook.xml"/>
                </Relationships>
                """),
            ("xl/workbook.xml",
                """
                <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
                <workbook xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" xmlns:r="http://schemas.openxmlformats.org/officeDocument/2006/relationships">
                  <sheets><sheet name="Sheet1" sheetId="1" r:id="rId1"/></sheets>
                </workbook>
                """),
            ("xl/_rels/workbook.xml.rels",
                """
                <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
                <Relationships xmlns="http://schemas.openxmlformats.org/package/2006/relationships">
                  <Relationship Id="rId1" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/worksheet" Target="worksheets/sheet1.xml"/>
                  <Relationship Id="rId2" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/sharedStrings" Target="sharedStrings.xml"/>
                  <Relationship Id="rId3" Type="http://schemas.openxmlformats.org/officeDocument/2006/relationships/styles" Target="styles.xml"/>
                </Relationships>
                """),
            ("xl/styles.xml",
                """
                <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
                <styleSheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
                  <fonts count="1"><font><sz val="11"/><name val="Calibri"/></font></fonts>
                  <fills count="1"><fill><patternFill patternType="none"/></fill></fills>
                  <borders count="1"><border/></borders>
                  <cellStyleXfs count="1"><xf numFmtId="0" fontId="0" fillId="0" borderId="0"/></cellStyleXfs>
                  <cellXfs count="1"><xf numFmtId="0" fontId="0" fillId="0" borderId="0" xfId="0"/></cellXfs>
                </styleSheet>
                """),
            ("xl/sharedStrings.xml", SharedStrings),
            ("xl/worksheets/sheet1.xml",
                """
                <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
                <worksheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
                  <sheetData>
                    <row r="1"><c r="A1" t="s"><v>0</v></c></row>
                    <row r="2"><c r="A2" t="s"><v>1</v></c></row>
                  </sheetData>
                </worksheet>
                """),
        };

        using var ms = new MemoryStream();
        using (var zip = new ZipArchive(ms, ZipArchiveMode.Create, leaveOpen: true))
        {
            foreach (var (path, content) in parts)
            {
                var e = zip.CreateEntry(path, CompressionLevel.Optimal);
                using var w = new StreamWriter(e.Open(), new UTF8Encoding(false));
                w.Write(content.TrimStart());
            }
        }
        return ms.ToArray();
    }
}
```

### Expected Behavior

Runs that had no `<color>` in the source should round-trip with no `<color>` element (or otherwise remain "automatic"), so Excel keeps rendering them with the automatic/theme colour and conditional formatting / theme colour changes still take effect. The phonetic-only string should stay a plain `<t>` + `<rPh>` and not be promoted into a coloured run.

### Actual Behavior

Every colour-less run is written with `<color rgb="FF000000"/>`, and the phonetic-only string is promoted to a rich-text run that also gets `rgb="FF000000"`.

Output `sharedStrings.xml` from the test above (XLibur 0.106.0):

```xml
<x:sst xmlns:x="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
  <x:si>
    <x:r>
      <x:rPr>
        <x:b />
        <x:vertAlign val="baseline" />
        <x:sz val="11" />
        <x:color rgb="FF000000" />   <!-- injected -->
        <x:rFont val="Calibri" />
        <x:family val="2" />
      </x:rPr>
      <x:t>Bold</x:t>
    </x:r>
    <x:r>
      <x:rPr>
        <x:vertAlign val="baseline" />
        <x:sz val="11" />
        <x:color rgb="FF000000" />   <!-- injected -->
        <x:rFont val="Calibri" />
        <x:family val="2" />
      </x:rPr>
      <x:t xml:space="preserve"> plain</x:t>
    </x:r>
  </x:si>
  <x:si>
    <x:r>
      <x:rPr>
        <x:vertAlign val="baseline" />
        <x:sz val="11" />
        <x:color rgb="FF000000" />   <!-- injected; string was plain text + phonetics -->
        <x:rFont val="Calibri" />
        <x:family val="2" />
      </x:rPr>
      <x:t>漢字</x:t>
    </x:r>
    <x:rPh sb="0" eb="2"><x:t>かんじ</x:t></x:rPh>
    <x:phoneticPr fontId="0" type="Hiragana" />
  </x:si>
</x:sst>
```

### Workaround

None at the model level — the writer always materialises the font. Consumers that must preserve fidelity currently have to post-process the saved package to strip `<color rgb="FF000000"/>` from runs they know were originally colour-less, which is fragile.

The underlying cause (inherited from the ClosedXML lineage) is that a run's colour is deserialized into a non-nullable `XLColor` that defaults to black, so "unset" and "explicit black" become indistinguishable by save time; and there is no non-rich representation for a `<t>` + `<rPh>` string, so it is wrapped in a synthetic run. A real fix likely needs the run-font colour to carry an "explicitly set" flag (tri-state / nullable) and the reader to preserve the phonetic-only shape.

### Willing to contribute a fix?

- [ ] I would be willing to submit a PR to fix this issue

jafin:
Fixes #219.

A plain load → `SaveAs()` round-trip wrote every colour-less rich-text run back with an explicit `<color rgb="FF000000"/>`, and promoted a plain string carrying a phonetic guide (`<t>` + `<rPh>`) into a synthetic run that inherited the same injection. Both silently alter the source formatting — an explicit black can't be overridden by a theme colour change or by conditional formatting, where the automatic colour it replaced can.

## Root cause

Two separate losses of "was this ever set?" information:

1. **Run colour.** A run's font is seeded from the cell font, whose colour defaults to black, and `TextSerializer.WriteRun` emitted `<color>` unconditionally.
2. **Phonetic-only strings.** `LoadPhonetics` reaches phonetics through `XLCell.GetRichText()`, which materialises the text into a run.

## Changes

- **`XLColor.IsUnset`** names the existing `NoColor` sentinel. It deliberately does *not* match indexed colour 64 the way `IsTransparent` does — that's an explicitly written value and must round-trip as such.
- **Reader** (`SetCellText`): runs are seeded from a cell font with the colour stripped, so an `rPr` without `<color>` stays automatic; `LoadFont` puts a colour back when the `rPr` carries one. Seeding rather than recolouring after `AddText` matters — every run mutation re-interns the whole rich text in the shared string table, and the extra churn reshuffled SST ids.
- **Writer**: `<color>` is omitted for an unset colour; a runless rich text writes a bare `<t>`.
- **Reader**: after loading a runless string's phonetics, the synthetic run is dropped again.

## Judgement calls

- **A run with no `<rPr>` at all** inherits the cell font per ECMA-376 CT_RElt, so that inheritance is kept — *including* an explicit cell font colour (covered by `RunWithoutRunProperties_StillInheritsAnExplicitCellFontColor`). The colour is dropped only when it is the ambiguous black default. Fully correct handling would mean recording "this run had no rPr" and writing none back, which needs a flag threaded through `XLRichString`/`RichTextRun` — much more invasive, and worth doing separately if wanted.
- **Style fonts are untouched.** A `<font>` in `styles.xml` with no `<color>` still loads as black. Fixing that root cause means making `XLFontValue.Default.FontColor` unset, which changes `cell.Style.Font.FontColor` for every default-styled cell — a public behaviour change beyond this bug, so it's left for a separate PR.
- **`XLImmutableRichText.Runs` can now be empty while `Text` is not.** Two consumers needed updating: `XLCellGlyphHelper` (column sizing measured such a cell as empty) and the `XLRichText` copy ctor (materialises a run on access, so the run-based mutable API doesn't write back an empty cell on the first edit). Both are covered by tests.

## Tests

New `RichTextColorRoundTripTests` builds minimal spec-valid packages in memory (the input can't be produced through the `XLWorkbook` API — the writer would normalise it before it were ever saved):

- colour-less runs don't gain an explicit black
- a phonetic-only string isn't promoted to a run
- an explicitly written run colour still survives
- a run without `rPr` keeps an explicit cell font colour
- editing a phonetic-only cell materialises a run without losing text or phonetics
- a phonetic-only cell is still measured by `AdjustToContents`

`kanji-with-new-line-output.xlsx` encoded the old promotion and was regenerated. Its input is exactly `<t>` + `<rPh>` + `<phoneticPr>` and the new output matches that shape — only `xl/sharedStrings.xml` changed, and it still passes the OpenXML SDK validator (`validate: true`). The `xml:space="preserve"` that test exists to protect is intact.

Full suite green on net8.0 and net10.0 (6340 passed); Release solution build clean under `TreatWarningsAsErrors`.


SamPlusPlus:
🤖 **Bot-assisted review note** (not a blocking request)

**Consideration:** `XLRichText` copy ctor materialization of a runless (phonetic-only) string

`GetRichText()` alone does not write back — `_cell` is still null during the ctor, so a plain load → save stays clean. The residual case is only the **first edit** of a phonetic-only cell:

- Today, `AddText(original.Text)` seeds from `cell.Style.Font`, which still carries default black.
- On the first mutation, that run is persisted, and the writer emits `<color rgb="FF000000"/>` because black is not `IsUnset`.
- Load already avoids this for real runs via `ColorlessFont` in `SetCellText` (clone cell font, set `FontColor = NoColor`).

**Optional consistency:** apply the same heuristic when materializing the synthetic run in the copy ctor — if the cell font color is the default black, seed with `NoColor` (automatic); keep an explicit cell color as-is. Then the first edit would omit `<color>` rather than pin black.

Out of scope for the #219 round-trip fix; only matters if someone opens a phonetic-only cell via `GetRichText()` and then mutates it.

XLibur/Excel/IO/WorksheetSheetDataReader.cs
```diff
@@ -723,17 +723,33 @@ internal static void SetCellText(XLCell xlCell, RstType element)
...
 
             if (runProperties == null)
-                xlCell.GetRichText().AddText(text, xlCell.Style.Font);
+            {
+                // No rPr at all: the run inherits the cell font (ECMA-376 CT_RElt), so materializing
+                // that font is faithful - except for a color the cell font only carries because it
+                // was itself defaulted to black, which would invent a color the source never had.
+                var cellFont = xlCell.Style.Font;
+                var font = cellFont.FontColor.Key == XLFontValue.Default.Key.FontColor
+                    ? ColorlessFont(cellFont, ref colorlessCellFont)
```
SamPlusPlus, line 740: @jafin I don't think this should be blocking, but all these references to ColorLessFont, but I think the better long term solution is to match ClosedXML and properly model Automatic Color https://github.com/ClosedXML/ClosedXML/commit/d9391901a2eaa350e93bce2afc5025b97913ba56 and https://github.com/ClosedXML/ClosedXML/commit/ea25deef8b918441691341b3f769b72fb6116c37

jafin:
Pushed a1c64f2b — a defect in this PR's own runless `<t>` writer, found in a review pass.

A runless shared string (plain text + phonetic guide) is decoded with `XmlEncoder.DecodeString` on the way in, but the new writer emitted the decoded text raw. A decoded `_xHHHH_` escape is a literal control character, which isn't valid XML content, so saving one threw:

```
System.ArgumentException: '\u0018', hexadecimal value 0x18, is an invalid character.
```

Fixed by encoding in the runless branch only. Run text must keep going out raw — it's never decoded on the way in, so encoding there would corrupt text that merely *looks* like an escape (`Hello_x0041_World`). This now matches what `SharedStringTableWriter` already does for plain shared strings.

The promoted-run path had the same defect before this branch, so it's latent rather than a regression — but it now lives on the code path this PR introduces, and the covering test (`PhoneticOnlyString_RoundTripsEscapedControlCharacters`) fails without the fix.

Suite green on net8.0 and net10.0 (6341 passed); Release build clean under `TreatWarningsAsErrors`.

Note for #227, which is stacked on this branch: it carries the same fix plus a second one (`Substring` dropping the no-rPr flag). When this merges, #227 will need a rebase — the `TextSerializer` change is identical in both, so it should resolve cleanly.

XLibur/Excel/IO/WorksheetSheetDataReader.cs
jafin: Thanks, I'll take a look at what's involved in aligning closer to ClosedXML enum.  Likely will do as another PR to not conflate this one.  Hopefully just naming conventions.

PR #227, opened 2026-07-26
fix: write runs that had no rPr back without one (#219)

jafin:
> **Stacked on #225** — it contains that PR's commit as well. Please merge #225 first; this PR's diff will then reduce to its own single commit. Addresses the first judgement call listed in #225.

## Problem

A rich-text run read from an `<r>` with no `<rPr>` inherits the cell font per ECMA-376 CT_RElt. The loader materialised that inherited font onto the run, so the save wrote it back as an explicit `<rPr>` — turning inheritance into formatting the source never stated, and dragging the cell font's colour along with it.

#225 worked around this by dropping the run colour when it was the ambiguous black default. **That heuristic is removed here.** The run now records that it states no formatting of its own, and the writer emits `<r><t>..</t></r>` with no `rPr` at all. The result round-trips exactly, and a cell font with an explicit colour is no longer a special case.

## Changes

- `XLRichString.InheritsContainerFont` marks a run as carrying no formatting of its own. Every font setter now goes through `OnFontChanged`, which clears it — once the caller sets any font property the formatting is the run's own and a full `rPr` is written. A text-only change does not clear it.
- `XLFormattedText.AddInheritedText` is the loader's entry point. Runs created through the public API are never flagged, so **workbooks built with XLibur are unaffected**.
- The flag is carried on `XLImmutableRichText.RichTextRun` and participates in equality and hashing, so two shared strings that differ only in `rPr` presence don't dedupe into one.

The inherited font is still held on the run, so reading and measuring (`XLCellGlyphHelper`, `cell.GetRichText()`) are unchanged. Comment runs load through `XLComment.AddText` and are never flagged, so comment output is unchanged.

No public API change.

## Tests

`RunWithoutRunProperties_StillInheritsAnExplicitCellFontColor` from #225 asserted the heuristic and is replaced by three tests, all against a workbook whose cell font is explicitly green (`FF00FF00`) with an `<r>` that has no `rPr`:

- `RunWithoutRunProperties_IsWrittenBackWithoutRunProperties` — no `rPr` appears in the output
- `RunWithoutRunProperties_StillReadsTheInheritedCellFont` — the run still reports the green colour in the object model
- `EditingRunWithoutRunProperties_MakesTheFormattingItsOwn` — after `SetBold()` the run writes a full `rPr` including the materialised green, so the edit isn't dropped

No reference fixtures needed regenerating. Full suite green on net8.0 and net10.0 (6342 passed); Release solution build clean under `TreatWarningsAsErrors`.

### DIFF
--- a/XLibur.Tests/Excel/RichText/RichTextColorRoundTripTests.cs
+++ b/XLibur.Tests/Excel/RichText/RichTextColorRoundTripTests.cs
@@ -122,35 +122,78 @@ public class RichTextColorRoundTripTests
         });
     }
 
+    // A run with no rPr states no formatting of its own; per ECMA-376 CT_RElt it inherits the cell
+    // font. The cell font here is explicitly green, so nothing about this run is ambiguous - the
+    // only faithful output is a run with no rPr at all.
+    private const string GreenFontStyles =
+        """
+        <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+        <styleSheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
+          <fonts count="1"><font><sz val="11"/><color rgb="FF00FF00"/><name val="Calibri"/></font></fonts>
+          <fills count="1"><fill><patternFill patternType="none"/></fill></fills>
+          <borders count="1"><border/></borders>
+          <cellStyleXfs count="1"><xf numFmtId="0" fontId="0" fillId="0" borderId="0"/></cellStyleXfs>
+          <cellXfs count="1"><xf numFmtId="0" fontId="0" fillId="0" borderId="0" xfId="0" applyFont="1"/></cellXfs>
+        </styleSheet>
+        """;
+
+    private const string BareRunSharedStrings =
+        """
+        <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+        <sst xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main" count="1" uniqueCount="1">
+          <si><r><t>Inherited</t></r></si>
+        </sst>
+        """;
+
     [Test]
-    public void RunWithoutRunProperties_StillInheritsAnExplicitCellFontColor()
+    public void RunWithoutRunProperties_IsWrittenBackWithoutRunProperties()
     {
-        // A run with no rPr inherits the cell font (ECMA-376 CT_RElt). Dropping the color is only
-        // correct when it is the ambiguous black default - an explicitly styled cell font must
-        // still be materialized onto the run.
-        const string greenFontStyles =
-            """
-            <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
-            <styleSheet xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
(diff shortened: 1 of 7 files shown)


jafin:
Pushed 722609e9 — two defects found in a review pass over this stack, both with tests that fail before the fix.

**1. Saving a runless string with an escaped control character threw.** A runless shared string (plain text + phonetic guide) is decoded with `XmlEncoder.DecodeString` on the way in, but the new `<t>` writer emitted the decoded text raw. A decoded `_xHHHH_` escape is a literal control character, which isn't valid XML content:

```
System.ArgumentException: '\u0018', hexadecimal value 0x18, is an invalid character.
```

Fixed by encoding in the runless branch only. Run text must keep going out raw — it's never decoded on the way in, so encoding there would corrupt text that merely *looks* like an escape. This now matches what `SharedStringTableWriter` already does for plain shared strings. The promoted-run path had the same defect before this stack, so it's latent rather than a regression, but it now lives on the new code path.

**2. `Substring` re-materialised the ambiguous black.** `XLFormattedText.Substring` rebuilt each piece of a split run without carrying `InheritsContainerFont`, so splitting a run that stated no formatting wrote the inherited cell font as an explicit `rPr` — putting `rgb="FF000000"` straight back for a default-styled cell. A split only divides text; every piece keeps the same font, so the flag has to survive. This *was* a regression against #225, which seeded such runs colourless.

Deliberately **not** propagated through `CopyFrom` or the copy constructor: those move a run to a different container, where "inherits" means picking up a different font, and silently changing appearance would be worse than materialising it.

Suite green on net8.0 and net10.0 (6344 passed); Release build clean under `TreatWarningsAsErrors`.

jafin:
Rebased onto `main` now that #225 has merged (`d8276153`). Force-pushed `10b47d85`.

#225 was **squash-merged**, so its commit on this branch couldn't be matched automatically — I rebased with `--onto main 8cf46964` to drop it and replay only this PR's own two commits. The PR now shows just those, and GitHub reports it mergeable.

One thing the rebase got wrong that's worth flagging, since it applied without reporting a conflict: `PhoneticOnlyString_RoundTripsEscapedControlCharacters` ended up **duplicated** in `RichTextColorRoundTripTests.cs`. It reached `main` via #225's encoding fix and also exists in this branch's `722609e9`, and the two copies landed at different offsets, so git merged them textually rather than flagging an overlap. Two identical methods in one class doesn't compile. Removed the second copy; the surviving one is byte-identical.

The encoding fix in `TextSerializer` was correctly recognised as already applied and is present exactly once — verified.

Suite green on net8.0 and net10.0 (6434 passed, now including main's chart tests); Release build clean under `TreatWarningsAsErrors`.

Still open here, unchanged by the rebase: the `colorlessCellFont` / `ColorlessFont` naming in `WorksheetSheetDataReader` and the `XLColor.IsUnset` helper — the subject of the review comment. #232 adds `XLColorType.Automatic` and `XLColor.Automatic` as the spec-aligned vocabulary for exactly this; once that lands, this PR can drop `IsUnset` and rename the helper to match. Happy to fold that in here instead if you'd rather not wait on #232.

PR #244, opened 2026-07-26
fix: position row-only and column-only references on deletion

jafin:
## Summary

Follow-up to #243, now merged; this targets `main` and contains one commit.

#243 noted a gap it deliberately left alone: a row-only reference (`Sheet1!$3:$4`) whose rows are removed by an exact-cover delete landed on `Sheet1!$1:$2` instead of `#REF!`. Looking properly, that symptom was the least of it — the row-only branch shifted both endpoints and clamped, without ever consulting where the deletion started. The cell-range branch received the top-boundary rules in #2866 and the `#REF!` invalidation in #243; this branch received neither.

Three distinct cases were wrong, on both axes:

| reference | deletion | before | correct |
| --- | --- | --- | --- |
| `$3:$5` | row 4 (inside) | `$2:$4` | `$3:$4` |
| `$3:$5` | row 3 (top boundary) | `$2:$4` | `$3:$4` |
| `$3:$4` | rows 3:4 (exact cover) | `$1:$2` | `#REF!` |
| `$C:$E` | column C | `$B:$D` | `$C:$D` |
| `$B:$E` | columns C:D | `$A:$C` | `$B:$C` |
| `$C:$D` | columns C:D | `$A:$B` | `#REF!` |

The first two are silent data corruption: the reference walks onto rows it never covered and drops rows that survived.

Insertion was wrong for the same reason, since it runs through the same branch:

| reference | insertion | before | correct |
| --- | --- | --- | --- |
| `$3:$5` | 2 rows at row 4 | `$5:$7` | `$3:$7` |
| `$C:$E` | 2 columns before D | `$E:$G` | `$C:$G` |

The equivalent cell ranges (`$A$3:$A$5` → `$A$3:$A$7`) have always expanded correctly; only the row-only and column-only forms moved off the rows they covered.

The exact-cover case is a consequence of the other two. `IXLRows.Delete()` applies a multi-row delete one row at a time, so a reference that drifts upward instead of shrinking never gets small enough for #243's invalidation check to fire.

## Changes

`XLibur/Excel/Cells/XLCellFormulaShifter.cs`

- `AppendShiftedRowOnlyRange` now decides both boundaries the way the cell-range branch does: the top only moves when the deletion starts at or above it, and stays pinned at the deletion start when the deletion removes it (`IsTopBoundaryDeletion`, shared with the cell-range path); the bottom always shifts. Rendering stays row-only and preserves whatever `$` the reference carried.
- `AppendShiftedColumnOnlyRange` had the identical three defects and gets the same treatment, sharing `IsLeftBoundaryDeletion`.
- `ShiftRowString` / `ShiftColumnString` are replaced by `AppendRowBoundary` / `AppendColumnBoundary`, which render one already-decided boundary rather than folding the shift arithmetic into the string handling.
- Both helpers name the boundary `shiftStart`, not `deletionStart`: insertions reach the same code.

## Related Issues

Completes the row and column deletion half of ClosedXML/ClosedXML#880, and extends the ClosedXML/ClosedXML#2866 top-boundary rules to the row-only and column-only references they missed.

## Testing

- [x] All existing tests pass
- [ ] Tested manually

Ten cases added to `NamedRangeDeletionTests`. For each axis: the exact cover, the boundary-removing delete, and the delete inside — plus the two cases that were already correct (a deletion entirely before the reference, and one overshooting its top, which reached the right answer only because clamping happened to cancel the drift) so the new positioning cannot regress them. All ten fail without this change and pass with it.

Four more in `NamedRangeInsertionTests`, alongside the cell-range cases they mirror: inserting inside and inserting before, for both axes.

Full suite on net8.0 and net10.0: 6508 total, 6504 passed, 0 failed, 4 skipped.

## Checklist

### DIFF
--- a/XLibur.Tests/Excel/NamedRanges/NamedRangeDeletionTests.cs
+++ b/XLibur.Tests/Excel/NamedRanges/NamedRangeDeletionTests.cs
@@ -183,6 +183,163 @@ public class NamedRangeDeletionTests
     }
 
     /// <summary>
+    /// Row-only counterpart of DeletingExactlyTheRowsOfNamedRange_BecomesRefError. A multi-row delete is
+    /// applied one row at a time, so the reference has to shrink correctly on the way down for the last
+    /// pass to see it as fully deleted.
+    /// </summary>
+    [Test]
+    public async Task DeletingExactlyTheRowsOfRowOnlyNamedRange_BecomesRefError()
+    {
+        using var wb = new XLWorkbook();
+        var ws = wb.AddWorksheet("Sheet1");
+        wb.DefinedNames.Add("Rows", "Sheet1!$3:$4");
+
+        ws.Rows(3, 4).Delete();
+
+        await Assert.That(RefersTo(wb, "Rows")).IsEqualTo("Sheet1!#REF!");
+    }
+
+    /// <summary>
+    /// Row-only counterpart of the #2866 top-boundary shrink: rows 3:5 with row 3 deleted becomes 3:4, not
+    /// 2:4. The row-only branch shifted both endpoints regardless of where the deletion started.
+    /// </summary>
+    [Test]
+    public async Task DeletingTopRowOfRowOnlyNamedRange_ShrinksAndShifts()
+    {
+        using var wb = new XLWorkbook();
+        var ws = wb.AddWorksheet("Sheet1");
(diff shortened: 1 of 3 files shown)


jafin:
@coderabbitai review

jafin:
@coderabbitai review

PR #250, opened 2026-07-26
docs: add the changelog-generator skill and project MCP config

jafin:
Checks in the Claude Code skill that generates and maintains `CHANGELOG.md`, so the conventions it encodes live in the repo rather than in one machine's `~/.claude`.

Independent of #249 — no overlapping files, and either can merge first.

## What the skill encodes

These are the conventions the `Unreleased` section in #249 now follows, written down so the next run reproduces them rather than reinventing a format:

- **Two-level grouping.** Change type first as emoji headings (`⚠️ Breaking Changes`, `✨ New Features`, `⚡ Performance`, `🐛 Bug Fixes`, `🗑️ Deprecations`), then feature area within each. A reader looking for what breaks reads one section; a reader who only cares about charting scans one subheading. Sixteen bug fixes in a flat list is not a changelog.
- **Per-entry attribution** — PR link plus author on every entry. Includes the two things that silently produce wrong output: PRs live on `XLibur/XLibur` rather than the `origin` fork, so linking `origin` yields 404s; and a squash merge rewrites the git author, so `%an` can name the merger rather than the contributor. `gh pr view --json author` is the source of truth for both.
- **Before/after examples** for anything requiring the reader to edit code, with the rule that names come from reading the diff rather than from memory — a plausible-looking wrong example is worse than no example. Plus when to skip it, and when to label a migration as non-mechanical.
- **A `Contents` table** linking each version including `Unreleased`, with GitHub's anchor-slug derivation spelled out and worked through, because a hand-guessed anchor fails silently (`## v0.106.0 - 2026-07-25` → `#v01060---2026-07-25`).
- **Leave released sections alone** when the format changes. Mixed heading styles across releases are expected; backfilling means mapping every historical entry to a PR, which is its own job.

It also keeps the existing noise filter (Sonar/style commits, test-only changes, CI config, internal renames) and the review checklist, which now flags invented feature areas, unattributable entries, and unverified migration examples.

## Two things worth a maintainer decision

- **`.mcp.json` rides along in the same commit.** It declares the `roslyn-refactor` MCP server so the project's tooling config travels with the repo instead of being set up per clone. It's unrelated to the skill beyond both being agent configuration — say the word and I'll split it into its own PR.
- **`.claude/` is in `.gitignore`** (line 33), so `SKILL.md` is force-added. That works, and once tracked its later edits show in `git status` normally. But if the intent is for agent config to be shared, the ignore rule is now half-true — worth deciding whether `.claude/skills/` should be un-ignored explicitly. There's also an untracked `.github/skills/` in the working tree that may be the intended home instead.

## Verification

Documentation and configuration only — no library code, no build or test impact.

### DIFF
--- a/.claude/skills/changelog-generator/SKILL.md
+++ b/.claude/skills/changelog-generator/SKILL.md
@@ -1,0 +1,356 @@
+---
+name: changelog-generator
+description: Generate or update CHANGELOG.md files from git history. Use this skill whenever the user asks to create a changelog, generate release notes, summarize commits for a release, document changes since the last version/tag, or update an existing CHANGELOG.md — even if phrased informally like "what changed since v2.4", "write up the release notes", or "add the recent commits to the changelog". Also trigger when the user mentions "Keep a Changelog", version release documentation, or wants commit history translated into user-facing language.
+---
+ 
+# Changelog Generator
+ 
+Generate and maintain professional, user-facing `CHANGELOG.md` files by analyzing git history, categorizing changes, and translating developer commits into customer language.
+ 
+## Workflow Overview
+ 
+1. **Determine the range** — figure out which commits to include
+2. **Scan git history** — collect commits with full context, plus author and PR for each
+3. **Filter noise** — exclude internal-only changes
+4. **Categorize** — group by change type (emoji sections), then by feature area within each
+5. **Translate** — rewrite commit messages as user-facing entries, each attributed to its PR and author
+6. **Format & write** — create or update `CHANGELOG.md` following the conventions below
+7. **Review with the user** — show the draft before finalizing
+## Step 1: Determine the Commit Range
+ 
+Resolve what the user means by "since last release" / "for version X" / "past week":
+ 
+```bash
+# Most recent tag (usually the last release)
+git describe --tags --abbrev=0
+ 
+# All tags, newest first, with dates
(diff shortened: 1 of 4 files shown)


jafin:
Pushed `4d799aea`, which addresses the `.mcp.json` point raised above rather than splitting it out.

**`roslyn-mcp` is now a repo tool, not a global one.** The config invoked the bare `roslyn-mcp` command, which resolves off `PATH` — so it only worked on a machine that happened to have `roslynmcp.server` installed globally. Anyone else cloning got a `roslyn-refactor` server that silently failed to start, with nothing in the repo naming the package or the expected version.

- `roslynmcp.server` 0.4.0 added to `.config/dotnet-tools.json` beside `dotnet-stryker`, pinned with `rollForward: false` to match.
- `.mcp.json` now runs `dotnet tool run roslyn-mcp`, resolving through the manifest.
- Installed via `dotnet tool install --local` rather than hand-editing the manifest, so the `commands` array comes from the package instead of a guess.
- `CONTRIBUTING.md` gains a **Local tools** section listing both tools and what each is for.

The trade is that the server now needs `dotnet tool restore` before first use. That failure is explicit and says what to do, unlike the old missing-command one, and CONTRIBUTING names it as the first thing to check — but it is a new first-run step worth knowing about.

Verified locally: `dotnet tool restore` succeeds, and `dotnet tool run roslyn-mcp` starts the server and speaks JSON-RPC over stdio.

PR #267, opened 2026-07-29
perf: find comment threads through the misc slice, not a full-sheet cell scan

jafin:
## Summary

`GeneratePersonPart`, added by #258, runs on every save of every workbook to decide whether
`xl/persons/person.xml` is needed. It answered that with

```csharp
CellsCollection.GetCells(c => c.HasThreadedComment)
```

which materialises an `XLCell` wrapper for every used cell in every worksheet to read one bool.
At 48 bytes a wrapper that is 24 MB on a 500K-cell save, paid in full by the overwhelming
majority of workbooks, which contain no threads at all.

It regressed the save phase of the 50K-row benchmark by 19%. Bisected across 8 build/measure
points: good at `1d07c076` (117.6 MB), bad at `e45e03d6` (140.5 MB).

| Scenario | Save before | Save after |
|---|---:|---:|
| `CreateAndSave` | 41.9 MB | **34.9 MB** |
| `CreateFormattedAndSave` | 140.5 MB | **117.6 MB** |

`dotnet run -c Release --project XLibur.Benchmarks --framework net10.0 -- profile alloc`, two runs
per side, stable to ±0.2 MB. Both figures are back to exactly what #185 recorded (34.9 / 117.6),
so this is a restoration rather than a new win. Create phase (183.3 MB) and wall time are
unchanged.

## Changes

- Add `XLCellsCollection.GetThreadedComments()`, which walks the misc slice the way the existing
  `HasAnyThreadedComment` and `GetCellImagePoints` already do, returning each thread paired with
  its `Point`.
- Route `PersonPartWriter.CollectReferencedPersons` through it. This is the call site that runs
  unconditionally, and the whole of the regression.
- Route `ThreadedCommentPartWriter.GenerateContent` through it too. That one is gated behind
  `HasAnyThreadedComment`, so it was not part of the regression, but it paid the same
  per-used-cell cost to write a handful of comments.

No public API change. No CHANGELOG entry: #258 has not shipped, so the regression never reached a
release, and the existing "50% fewer allocations on save" bullet is true again.

## Related Issues

Regression introduced by #258. This is the fifth instance of the pattern
[spec 03](https://github.com/XLibur/XLibur/blob/main/docs/specs/03-save-path-allocations.md) removed from four other save
helpers — see its Results section, "Full-sheet `GetCells()` scans".

## Testing

- [x] All existing tests pass
- [ ] Tested manually — not applicable; no behaviour change, and threaded-comment round trips are
      covered by the existing suite from #258.

New `ThreadedCommentSaveCostTests`:

- `Collecting_persons_does_not_materialise_a_cell_per_used_cell` asserts that collecting persons
  on a 20,000-cell sheet with no threads allocates under 64 KB. Verified it fails against the
  previous implementation, where it reports **960,784 bytes** — 20,000 × 48, plus change. That is
  what confirmed the attribution.
- `Threads_on_many_cells_round_trip_with_the_right_references` puts threads on three cells spread
  across rows and columns of a 20,000-cell sheet and round-trips them, pinning the row-major point
  pairing the new walk does.

Full suite green on both TFMs: 11,601 passed, 4 skipped (all pre-existing).

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved saving and reloading of threaded comments across multiple cells.
  * Preserved threaded comment text and author references during workbook round trips.
  * Reduced unnecessary processing when saving workbooks with many populated cells and few threaded comments.

* **Tests**
  * Added coverage for threaded comment persistence and save performance.

### DIFF
--- a/XLibur.Tests/Excel/Comments/ThreadedCommentSaveCostTests.cs
+++ b/XLibur.Tests/Excel/Comments/ThreadedCommentSaveCostTests.cs
@@ -1,0 +1,75 @@
+using System;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using XLibur.Excel;
+using XLibur.Excel.IO;
+
+namespace XLibur.Tests.Excel.Comments;
+
+/// <summary>
+/// The threaded-comment write path runs on every save of every workbook, including the
+/// overwhelming majority that have no threads at all. These tests pin down that it costs nothing
+/// on those workbooks — a scan that materialises an <c>XLCell</c> per used cell here is invisible
+/// in correctness tests but showed up as +23 MB on the 50K-row save benchmark.
+/// </summary>
+public class ThreadedCommentSaveCostTests
+{
+    private const int UsedCells = 20_000;
+
+    [Test]
+    public async Task Collecting_persons_does_not_materialise_a_cell_per_used_cell()
+    {
+        using var wb = new XLWorkbook();
+        var ws = wb.AddWorksheet("Sheet1");
+        for (var row = 1; row <= UsedCells; row++)
+            ws.Cell(row, 1).Value = row;
+
+        // Warm up so JIT and any first-call setup are outside the measured window.
+        PersonPartWriter.CollectReferencedPersons(wb);
+
+        var before = GC.GetTotalAllocatedBytes(precise: true);
+        var persons = PersonPartWriter.CollectReferencedPersons(wb);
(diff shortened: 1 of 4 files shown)


jafin:
@coderabbitai review

jafin:
@coderabbitai where was the first review?

jafin:
@coderabbitai so you cannot review this PR?

jafin:
@coderabbitai your cooked, you have not run a review in the last 9 minutes so the fair use policy seems incorrectly applied.

PR #287, opened 2026-07-31
refactor(pivot): move the save-time pivot cache id off XLPivotCache

jafin:
Third in the stack — **base is `fix/astnode-reference-cache`** (#286), which sits on `chore/todo-triage` (#285). Review #285 → #286 → this. Retarget as each merges.

Actions TODO item #12 from the triage: `PivotTableDefinitionPartWriter2.cs:79`, `xml.WriteAttribute("cacheId", pt.PivotCache.CacheId!.Value); // TODO: Maybe not nullable?`

## The answer to the TODO

It did not need to be nullable — but the reason is that it did not belong on the cache at all.

`XLPivotCache.CacheId` was documented as *"Pivot cache definition id from the file"*, which was never true. `PivotTableDefinitionPartReader.cs:233` **discards** the `cacheId` it reads:

```csharp
_ = pivotTable.CacheId?.Value ?? throw PartStructureException.MissingAttribute();
```

so nothing ever set the property on load. Its only writer is `SynchronizeWorkbookPivotCacheReferences`, which assigns each cache a position while rebuilding the workbook `pivotCaches` element, and its only reader is `PivotTableDefinitionPartWriter2`. The value is a property of *one save*, not of the cache — which is precisely why it had to be nullable, since a cache that has not been saved has no id.

## The change

Move it to `SaveContext` as a cache→id map behind a `GetPivotCacheId` accessor, and delete the property. The writer now reads a plain `uint`, and a lookup miss throws `UnreachableException` naming the invariant — the same shape as the existing `GetSharedStringId` — rather than the `NullReferenceException` you would get from dropping the `!`.

The ordering that makes this safe is unchanged, and is now stated where it matters: `CreateParts` runs `PreparePivotCaches` (registering every cache reachable from any pivot table on any sheet) before the per-sheet loop writes any pivot table part.

`PivotSourceCacheId` is gone from `SaveContext`. It was reset to `0` at the top of the only method that used it and read nowhere else — a loop counter with workbook-level scope. The loop now indexes directly.

**No behaviour change.** Ids are still positions in the rebuilt `pivotCaches` element, assigned in the same order.

## Tests

Two added to `XLPivotCacheTests`, both built on a workbook with **two distinct caches** — with only one cache a mix-up is invisible, whereas with two it points a pivot table at the wrong source rather than merely writing an odd number.

- `SavedPivotTables_ReferenceTheCacheIdTheWorkbookListsForTheirOwnCache` — every written `cacheId` matches a cache the workbook declares, and the two tables resolve to different ones.
- `SavingTwice_AssignsTheSameCacheIdsAgain` — ids are per-save, so a second save renumbers from scratch instead of continuing from the first.

Both were confirmed to have teeth: against a writer emitting a constant `cacheId` they fail with `Expected to be 2, but found 1`.

## Test plan

- [x] Full suite green on **net10.0** — 11,627 passed, 4 skipped
- [x] Full suite green on **net8.0** — 11,627 passed, 4 skipped
- [x] New tests verified to fail against a deliberately broken writer

### DIFF
--- a/XLibur.Tests/Excel/PivotTables/XLPivotCacheTests.cs
+++ b/XLibur.Tests/Excel/PivotTables/XLPivotCacheTests.cs
@@ -1,6 +1,11 @@
-﻿using XLibur.Excel;
+﻿using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Threading.Tasks;
+using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Spreadsheet;
 using TUnit.Assertions.Enums;
+using XLibur.Excel;
 
 namespace XLibur.Tests.Excel.PivotTables;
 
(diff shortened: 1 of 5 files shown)


jafin:
Code review flagged the `if (xlUsedCaches.Count != 0)` guard at `XLWorkbook_Save.cs:1031` — asking whether a save path could skip id assignment while still writing pivot table parts, which would now throw `UnreachableException` where the old code wrote a stale `CacheId`.

Confirmed it cannot:

- `xlUsedCaches` gets one entry per distinct `pt.PivotCache` over `allPivotTables`, so `Count == 0` ⟺ `allPivotTables` is empty (`pt.PivotCache` is non-nullable, and the first `HashSet.Add` always succeeds).
- `allPivotTables` (`XLWorkbook_Save.cs:676-681`) and the writer loop (`:200`) both iterate the same `WorksheetsInternal → ws.PivotTables`, unfiltered, and `DeleteRemovedWorksheets` has already run.
- `GeneratePivotTables` is itself guarded by `worksheet.PivotTables.Any<XLPivotTable>()`, so zero pivot tables means no part is written.
- `PreparePivotCaches` runs unconditionally at `:184`, before the per-sheet write loop.

So the guard is false only when there is nothing to write. Where the exception *is* a behaviour change is the case the old code handled badly: a cache that fell out of the workbook listing between saves previously wrote the **previous save's** id, silently pointing a pivot table at the wrong source. Failing loudly is the intended improvement.

No code change.

PR #290, opened 2026-07-31
fix(validation): never write a data validation with an empty sqref

jafin:
Actions TODO item #7 from the triage (`XLDataValidations.cs:278`, `//TODO Remove empty data validations`).

## The bug

A data validation rule applies to a set of ranges, written as the `sqref` attribute. The schema requires it to be non-empty, and Excel treats `sqref=""` as corruption — it repairs the workbook and drops **every** validation on the sheet, not just the broken one.

Two ways a rule ended up covering nothing, both of which reached the file:

**1. Adding a validation over a range that wholly contains an existing rule.** `SplitExistingRanges` carves the new range out of every overlapping rule via `SplitBy`, which removes each intersecting area and adds back only the remainder pieces (`XLDataValidation.cs:119-133`). When an area is fully covered, `Exclude` yields no pieces — so a rule whose entire coverage was inside the new range loses all of it. The rule stayed in the collection, `Consolidate` preserved it (it has no zero-area filter), and `DataValidationWriter.cs:102` joined its zero ranges into `""`.

```csharp
ws.Range("A1:A3").CreateDataValidation().WholeNumber.EqualOrGreaterThan(1);
ws.Range("A1:A5").CreateDataValidation().WholeNumber.EqualOrGreaterThan(2);
// before: two rules, saved as sqref="" and sqref="A1:A5"
```

**2. `ClearRanges` and `RemoveRange` are public** and neither drops the rule, so a caller can empty one directly with no splitting involved.

## The fix

Both ends:

- `SplitExistingRanges` deletes any rule left with no areas once the split completes. The new rule took every cell it applied to, so the old one is unreachable.
- `DataValidationWriter` skips any rule with no coverage, which closes (2) as well. Skipping loses nothing — a rule applying to no cell has nothing to say. When the emptied rule was the sheet's only one, the `dataValidations` element is now omitted entirely rather than written holding a single broken child.

**Why the sweep runs after the loop rather than off the coverage-changed event:** `SplitBy` drops a rule to zero areas *transiently*, between removing an area and adding back its remainder pieces. A rule with one area being split in two passes through zero on the way. Reacting to the event would delete a rule that is about to be repopulated.

Also deduped the split loop — the index holds one entry per area, so a rule covering several intersecting areas was split once per area. `SplitBy` is idempotent so this was harmless, but the emptied-rule sweep needs each rule considered once.

## Tests

6 added in `EmptyDataValidationTests`: the covered-rule case, the partial-overlap case that must still keep its remainder (`A1:A5` minus `A2:A3` → `A1`, `A4:A5`), both save paths, the only-rule-on-the-sheet case, and a round trip.

**4 of the 6 fail against the unfixed code** — verified by stashing the fix:

```
failed AddingValidationOverAnExistingOne_DropsTheRuleItFullyCovers   Expected to be 1, but found 2
failed ValidationCoveredByAnother_IsAbsentFromTheSavedFile           Expected to be equivalent to [A1:A5]
failed ValidationWithClearedRanges_IsAbsentFromTheSavedFile          Expected to be equivalent to [C1:C3]
failed SheetWhoseOnlyValidationWasCleared_WritesNoDataValidations    Expected to be false, but found True
```

The other two are behaviour-preserving regression guards.

## Test plan

- [x] Full suite green on **net10.0** — 11,620 passed, 4 skipped
- [x] Full suite green on **net8.0** — 11,620 passed, 4 skipped
- [x] New tests verified to fail on the unfixed code

### DIFF
--- a/XLibur.Tests/Excel/DataValidations/EmptyDataValidationTests.cs
+++ b/XLibur.Tests/Excel/DataValidations/EmptyDataValidationTests.cs
@@ -1,0 +1,155 @@
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Spreadsheet;
+using XLibur.Excel;
+
+namespace XLibur.Tests.Excel.DataValidations;
+
+/// <summary>
+/// A data validation rule applies to a set of ranges, written as the <c>sqref</c> attribute.
+/// The schema requires that attribute to be non-empty, so a rule that has ended up covering
+/// nothing must never reach the file: Excel treats <c>sqref=""</c> as corruption and repairs
+/// the workbook, dropping every validation on the sheet.
+/// </summary>
+public class EmptyDataValidationTests
+{
+    [Test]
+    public async Task AddingValidationOverAnExistingOne_DropsTheRuleItFullyCovers()
+    {
+        using var wb = new XLWorkbook();
+        var ws = wb.AddWorksheet("Data");
+
+        ws.Range("A1:A3").CreateDataValidation().WholeNumber.EqualOrGreaterThan(1);
+        await Assert.That(ws.DataValidations.Count()).IsEqualTo(1);
+
+        // Covers every cell the first rule applied to, so that rule now applies to nothing.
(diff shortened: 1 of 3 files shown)


jafin:
@coderabbitai force review

PR #291, opened 2026-07-31
feat(pivot): round-trip the chartFormats collection

jafin:
Actions TODO item #10 from the triage (`PivotTableDefinitionPartReader.cs:127`, `// TODO: chartFormats`).

**Base is `fix/pivot-cache-id-nullability`** (#287), because it touches the same writer. It is a *sibling* of #288, not stacked under it — both branch from #287 and are independent of each other.

## The data loss

A PivotChart keeps its manual per-series and per-point formatting in the chart part, but the `chartFormats` collection on the pivot table is what ties each formatting record to the pivot area it applies to. The reader skipped the element entirely, so loading and saving a workbook with a formatted PivotChart dropped every link and the formatting silently disappeared.

Worth separating from something that already worked: the `chartFormat` **attribute** was handled (`XLPivotTable.ChartFormat`, written at `PivotTableDefinitionPartWriter2.cs:143`) — that is the pivot table id `/chartSpace/pivotSource/fmtId/@val` points at. This PR is about the `chartFormats` **element**.

## The change

Follows the existing `formats`/`conditionalFormats` pattern exactly:

- `XLPivotChartFormat`, mirroring `XLPivotFormat`
- `LoadChartFormats` in the reader, `WriteChartFormats` in the writer
- both reuse `LoadPivotArea`/`WritePivotArea`, so the pivot area travels with the ids

The write sits between `conditionalFormats` and `pivotTableStyleInfo`, where the schema sequence puts it, and emits nothing when the collection is empty — an empty `chartFormats` element is not valid, and almost no pivot table has a chart. `chart` and `format` are required attributes so they are always written; `series` only when true.

## Tests

4 added in `PivotChartFormatTests`, on a minimal workbook whose pivot table carries two records — one formatting a whole series, one a single point:

- the ids survive
- **the pivot area inside the record survives** — a round trip that kept only the ids would still lose the link, so this is asserted separately
- the loaded model exposes them
- nothing is written for a pivot table without a chart

**3 of the 4 fail against the unfixed reader/writer** — verified by stashing the change. The fourth is the no-chart case, which passed before and must keep passing.

## Test plan

- [x] Full suite green on **net10.0** — 11,631 passed, 4 skipped
- [x] Full suite green on **net8.0** — 11,631 passed, 4 skipped
- [x] New tests verified to fail without the reader/writer changes

### DIFF
--- a/XLibur.Benchmarks/FormulaEvaluationBenchmarks.cs
+++ b/XLibur.Benchmarks/FormulaEvaluationBenchmarks.cs
@@ -1,0 +1,106 @@
+using BenchmarkDotNet.Attributes;
+using XLibur.Excel;
+using XLibur.Fonts.SixLabors.V1;
+
+namespace XLibur.Benchmarks;
+
+/// <summary>
+/// Isolates the cost of resolving a <c>ReferenceNode</c> to a range during formula
+/// evaluation. <c>RecalculateAllFormulas</c> marks every formula dirty and then
+/// recalculates, so each iteration re-resolves every reference and the measurement is
+/// repeatable.
+/// </summary>
+/// <remarks>
+/// The three shapes separate the two costs that <c>ReferenceNode.GetReference</c> used to
+/// pay on every evaluation — parsing the address string, and allocating a <c>Reference</c>:
+/// <list type="bullet">
+/// <item>
+/// <see cref="UniqueSameSheet"/> gives every row its own formula text, so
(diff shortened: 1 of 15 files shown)


jafin:
Superseded by #294. The base branch `fix/pivot-cache-id-nullability` (#287) has merged, so this is the same change rebased onto `main` — no content differences.

PR #300, opened 2026-07-31
feat(pivot): round-trip the filters collection

jafin:
Actions TODO item #11 from the triage — the last data-loss item (`PivotTableDefinitionPartReader.cs`, `// TODO: filters`). Based on `main`.

## The data loss

The `filters` element holds the label, value, date and top-N filters applied to pivot fields — what Excel offers from a field's dropdown. The reader skipped it entirely, so loading and saving a workbook **silently un-filtered every pivot table in it**. That changes what the workbook *shows*, not just what metadata it carries.

Not to be confused with the report-filter axis: that is `pageFields`, modelled by `XLPivotTable.Filters`, and was already supported. The new collection is `XLPivotTable.PivotFilters`, and both summaries say which is which because the names are easy to mix up.

## Two deliberate modelling choices

**`type` is kept as its raw `ST_PivotFilterType` token, not an enum.** There are ~40 values, XLibur acts on none of them, and a token round-trips a file written by a newer Excel using a value this build has never heard of. An enum would silently drop those.

**`autoFilter` is preserved verbatim rather than modelled.** It is required by the schema and is a sub-tree in its own right — `filterColumn` with any of `filters`, `top10`, `customFilters`, `dynamicFilter`, `colorFilter`, `iconFilter`. Modelling all of that would be a large new surface with nothing reading it.

XLibur has no API for creating or interpreting pivot filters, so the goal here is **fidelity**, and the attributes that *are* modelled (`fld`, `id`, `type`, `name`, `description`, `stringValue1/2`, `evalOrder`, `mpFld`, `iMeasureHier`, `iMeasureFld`) are the ones an API would need first. Preserving a sub-tree as raw XML follows the existing precedent in `SharedStringReader` and `ThreadedCommentPartWriter`.

If you would rather see `CT_AutoFilter` fully modelled, say so and I will do it as a follow-up — but it is a lot of type surface to add speculatively.

## Placement

The write sits **after `pivotTableStyleInfo`**, which is where the schema sequence puts `filters` — not next to `formats`, despite the name. It emits nothing when the collection is empty, since an empty `filters` element is invalid and most pivot tables have none.

## Tests

5 added in `PivotFilterTests`, on a workbook whose pivot table carries a caption filter holding a value list and a value filter holding a top-N:

- the attributes survive
- **the preserved `autoFilter` survives, for both shapes** — asserted separately, because a round trip keeping only the attributes would still lose what the filter actually does
- the loaded model exposes them
- nothing is written for a pivot table without filters

**4 of the 5 fail against the unfixed reader/writer** — verified by stashing the change.

## Test plan

- [x] Full suite green on **net10.0** — 11,654 passed, 4 skipped
- [x] Full suite green on **net8.0** — 11,654 passed, 4 skipped
- [x] New tests verified to fail without the reader/writer changes

## Summary by CodeRabbit

* **New Features**
  * Added support for loading and saving pivot-table filters, including label, value, date, and top-N criteria.
  * Preserved filter settings and AutoFilter criteria when pivot-table files are reopened and saved.
  * Exposed loaded pivot filters through the pivot-table object model.
  * Prevented empty filter elements from being written when no filters are configured.

### DIFF
--- a/XLibur.Tests/Excel/PivotTables/PivotFilterTests.cs
+++ b/XLibur.Tests/Excel/PivotTables/PivotFilterTests.cs
@@ -1,0 +1,366 @@
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Threading.Tasks;
+using DocumentFormat.OpenXml;
+using DocumentFormat.OpenXml.Packaging;
+using DocumentFormat.OpenXml.Spreadsheet;
+using XLibur.Excel;
+using XLibur.Excel.IO;
+
+namespace XLibur.Tests.Excel.PivotTables;
+
+/// <summary>
+/// The <c>filters</c> collection holds the label, value, date and top-N filters applied to pivot
+/// fields — what Excel offers from a field's dropdown. Dropping it on load silently un-filters
+/// the pivot table on the next save, which changes what the workbook shows.
+/// </summary>
+/// <remarks>
+/// Not the report-filter axis: that is <c>pageFields</c>, modelled by
+/// <c>XLPivotTable.Filters</c>, and was already supported.
+/// </remarks>
+public class PivotFilterTests
+{
+    [Test]
+    public async Task PivotFilters_SurviveARoundTrip()
+    {
+        using var input = new MemoryStream();
+        CreateWorkbookWithPivotFilters(input);
+        input.Position = 0;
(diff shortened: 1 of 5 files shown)


jafin:
@coderabbitai review

PR #307, opened 2026-08-01
fix(report): bind defined names by scope, the way Excel resolves them

### LINKED ISSUE #274: XLibur.Report: defined names are de-duplicated workbook-wide, so a sheet-scoped name can be silently dropped
Found while reviewing `feat/spec-12-report-templating` (spec 12, `XLibur.Report`). Not a regression — it is the behaviour the binder was written with — but it produces silently wrong output, so it needs a decision before the package ships.

## What happens

`RangeBinder.Resolve` de-duplicates defined names with a single workbook-wide, case-insensitive `HashSet`, and `EnumerateNames` yields workbook-scoped names before sheet-scoped ones:

```csharp
// XLibur.Report/Ranges/RangeBinder.cs
var seen = new HashSet<string>(StringComparer.OrdinalIgnoreCase);

foreach (var definedName in EnumerateNames(workbook))
{
    if (!seen.Add(definedName.Name))
    {
        continue;
    }
    ...
}

private static IEnumerable<IXLDefinedName> EnumerateNames(IXLWorkbook workbook)
{
    foreach (var name in workbook.DefinedNames.ValidNamedRanges())
        yield return name;

    foreach (var worksheet in workbook.Worksheets)
        foreach (var name in worksheet.DefinedNames.ValidNamedRanges())
            yield return name;
}
```

Two consequences:

1. **Two sheets each declaring a sheet-scoped name of the same name.** This is legal in Excel and a natural way to write a template with one section per sheet — `Q1!Items` and `Q2!Items`. Only the first binds. The second sheet's range never expands, so its `{{ item.* }}` placeholders fall through to the pre-expansion pass where `item` is not in scope, and the cells render blank. No `TemplateError` is recorded, so `XLGenerateResult.Errors` is empty and the report looks like it worked.

2. **A workbook-scoped name and a sheet-scoped name sharing a name.** Excel's rule is that the sheet-scoped one shadows the workbook-scoped one *on its own sheet*. Here the workbook-scoped one wins everywhere and the sheet-scoped one is dropped.

Reproduction for (1): two sheets, each with a sheet-scoped `Items` over a template block, one `AddVariable("Items", …)`. The first sheet generates; the second comes out blank with no error.

## Decision needed

How much of Excel's name scoping should the binder honour?

**Option A — key by scope, keep one variable per name.** Change `seen` to key on `(owning sheet, name)` so every sheet-scoped name binds, and have a sheet-scoped name suppress the workbook-scoped one *for that sheet only*. All ranges named `Items` bind to the same `Items` variable. Fixes both cases, matches Excel's shadowing, and is the smallest change. It does mean two ranges rendering the same collection, which for case (1) is exactly what a per-sheet template wants.

**Option B — A, plus sheet-qualified variable lookup.** As A, but a sheet-scoped name may also resolve to a per-sheet variable (`Q1_Items`, or a `Sheet!Name` form). More expressive, more surface, and it needs a syntax decision.

**Option C — leave the behaviour, report it.** Keep first-wins but add a `TemplateError` when a name is skipped because one of that name already bound. Cheapest, and stops the silent-blank-report failure mode, but leaves multi-sheet templates unable to repeat a section.

My inclination is **A**: it is what the code already implies it does, it is what Excel does, and it turns a silent wrong answer into a correct one without adding template syntax. Worth confirming before I change binding semantics, though — anything that changes which ranges bind is a behaviour change for templates authored against the current build.

## Notes

- `TryResolveItems` / `TryResolveValue` need no change either way; only the de-duplication key and the workbook-vs-sheet precedence move.
- Whichever option is chosen, a test with two sheets each declaring the same sheet-scoped name should pin it — there is currently none.

jafin:
Fixes #274. Option A, as agreed on the issue.

## What was wrong

`RangeBinder.Resolve` de-duplicated defined names with a single workbook-wide, case-insensitive `HashSet<string>`, so of two names sharing a name only the first ever bound. Two consequences, both silent:

1. **Two sheets each declaring a sheet-scoped name of the same name** — legal in Excel, and the natural way to write a template with one section per sheet. Only the first bound. The second sheet's placeholders fell through to the pre-expansion pass, where `item` is not in scope, and rendered blank. Nothing reached `ParsingErrors`, so the report looked like it had worked.
2. **A workbook-scoped name and a sheet-scoped name sharing a name** — the workbook-scoped one won everywhere, where Excel has the sheet-scoped one shadow it on its own sheet.

## What it does now

Names are gathered under Excel's scoping rather than de-duplicated. Every sheet-scoped name binds. A workbook-scoped name binds except where a sheet declares its own name of that name.

Excel decides shadowing by where a reference is *written*, and a defined name here is not read from anywhere — it **is** a range. So the sheet it covers stands in for the sheet a reference would be written on, and shadowing reaches that sheet alone:

```
Workbook  Items -> Q1!A5:C5        binds: Q1!A5:C5   (Q1 declares no Items)
Sheet Q2  Items -> Q2!A5:C5        binds: Q2!A5:C5

Workbook  Items -> Q2!A5:C5        drops: Q2!A5:C5   (shadowed on its own sheet)
Sheet Q2  Items -> Q2!A10:C10      binds: Q2!A10:C10
```

Legitimate shadowing is **silent**. It is what Excel does, not a template mistake, and `TemplateErrors` has no warning severity — anything recorded there reads as an error and would show up on every well-formed template using the pattern.

The name still selects the variable, so every range named `Items` binds the `Items` collection. For case (1) that is exactly what a per-sheet template wants; a template that needs two sheets bound to different data gives the ranges different names. Documented in `docs/report-templating.md`.

## Cost

`IsShadowed` is one dictionary probe per workbook-scoped name. `RefersTo` is only parsed for a name that some sheet actually declares, which is almost never — so the common path does no more work than before, and one `HashSet` allocation less.

## Tests

`XLibur.Report.Tests/Ranges/DefinedNameScopeTests.cs`, six tests: the issue's two cases, the workbook name still binding on sheets that do not shadow it, both scopes binding when they cover different sheets, the plain single-name case unaffected, and name comparison.

`ShadowingComparesNamesTheWayExcelDoes` drives `RangeBinder.Resolve` directly rather than a template, and the test says why: XLibur holds defined names case-insensitively, but `XLTemplate._variables` is `StringComparer.Ordinal`, so neither `ITEMS` nor `items` would find an `Items` variable and both blocks would come out unbound — the two answers would be indistinguishable through a template. That mismatch is its own problem and has been raised separately.

467/467 report tests pass; solution builds clean with `TreatWarningsAsErrors`.

### DIFF
--- a/XLibur.Report.Tests/Ranges/DefinedNameScopeTests.cs
+++ b/XLibur.Report.Tests/Ranges/DefinedNameScopeTests.cs
@@ -1,0 +1,186 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using XLibur.Excel;
+using XLibur.Report.Ranges;
+
+namespace XLibur.Report.Tests.Ranges;
+
+/// <summary>
+/// Which defined names bind when more than one of them shares a name.
+/// </summary>
+/// <remarks>
+/// Excel scopes a defined name either to the workbook or to one sheet, and the same name may be
+/// declared once per sheet on top of a workbook-wide one. Binding follows those rules: every
+/// sheet-scoped name binds, and a workbook-scoped name binds except on a sheet that declares its
+/// own. Every range named <c>Items</c> reads the <c>Items</c> variable, which is what a template
+/// repeating one section per sheet wants.
+/// </remarks>
+public class DefinedNameScopeTests
+{
+    private static List<SaleItem> ThreeItems() =>
+        new[] { ("Widget", 2, 5m), ("Gadget", 3, 10m), ("Doohickey", 1, 7.5m) }
+            .Select(r => new SaleItem
+            {
+                Product = r.Item1,
+                Quantity = r.Item2,
+                UnitPrice = r.Item3,
+                SoldOn = new DateTime(2026, 1, 1),
+            })
+            .ToList();
+
+    /// <summary>
+    /// Adds a one-row template block at <paramref name="firstRow"/> — the row after it is the
+    /// options row — and returns the range the two occupy.
+    /// </summary>
+    private static IXLRange TemplateBlock(IXLWorksheet sheet, int firstRow)
+    {
+        sheet.Cell(firstRow, 1).Value = "{{ item.Product }}";
+        return sheet.Range(firstRow, 1, firstRow + 1, 3);
+    }
+
+    private static XLGenerateResult Generate(IXLWorkbook workbook, object? items)
+    {
+        using var template = new XLTemplate(workbook);
+        template.AddVariable("Items", items);
+        return template.Generate();
+    }
+
+    /// <summary>
+    /// The case from issue #274: one section per sheet, each marked with a sheet-scoped name of the
+    /// same name. Both used to be de-duplicated down to one, and the second sheet came out blank
+    /// with nothing in <see cref="XLGenerateResult.ParsingErrors"/> to say why.
+    /// </summary>
+    [Test]
+    public async Task TheSameSheetScopedNameOnTwoSheetsBindsOnBoth()
+    {
+        using var workbook = new XLWorkbook();
+        var first = workbook.AddWorksheet("Q1");
+        var second = workbook.AddWorksheet("Q2");
+        first.DefinedNames.Add("Items", TemplateBlock(first, 1));
+        second.DefinedNames.Add("Items", TemplateBlock(second, 1));
+
+        var result = Generate(workbook, ThreeItems());
+
+        await Assert.That(result.ParsingErrors).IsEmpty();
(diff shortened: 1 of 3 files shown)


jafin:
The `build-and-test` failure here is not from this PR — `main` is red.

`ExcelFunctionBridgeTests.RecordingEngine` does not implement `IExpressionEngine.Culture`, so the Build step fails on both target frameworks and no tests run. #305 added `Culture` to the interface; #306 added `RecordingEngine` from a branch that predated it. Clean merge, no conflict, broken combination — `main` at `877b2462` fails on its own run.

#309 fixes it. This branch needs a rebase once that lands; nothing here changes.

PR #311, opened 2026-08-01
fix(names): resolve a structured reference in a defined name the way the calc engine does

jafin:
Actions the `// TODO: Temporary placeholder, extract range detection from CalculationVisitor` in `FormulaReferences.CollectRefsFactory.StructureReference`.

## The TODO was still live, and the work it asked for is already done

`StructuredReferenceResolver` was extracted at some point and is shared by `CalculationVisitor` (which evaluates the reference) and `DependenciesVisitor` (which registers it as a precedent). It handles every `#Headers` / `#Data` / `#Totals` / `#All` / `#ThisRow` combination, column spans, and reversed corners. `FormulaReferences` is the one caller that never got migrated, so this migrates it rather than reimplementing anything.

`FormulaReferences` is used only by `XLDefinedName`, so the blast radius is defined names whose `RefersTo` is a structured reference (`=Sales[Amount]`).

## What was wrong

The old code stored `(table, firstColumn)` and resolved by hand as `table.DataRange!.Column(column)` — the data column, whatever the reference actually said. Measured against a `Sales` table at `A1:C4` (headers `A1:C1`, data `A2:C3`, totals `A4:C4`):

| Formula | Before | After |
|---|---|---|
| `Sales[Amount]` | `B2:B3` | `B2:B3` (unchanged) |
| `Sales[[Amount]:[Tax]]` | `B2:B3` | `B2:C3` |
| `Sales[[#Headers],[Amount]]` | `B2:B3` | `B1:B1` |
| `Sales[[#Totals],[Amount]]` | `B2:B3` | `B4:B4` |
| `Sales[[#All],[Amount]]` | `B2:B3` | `B1:B4` |
| `Sales[#Data]` | *(no ranges)* | `A2:C3` |
| `Sales[#All]` | *(no ranges)* | `A1:C4` |
| `Sales[NoSuchColumn]` | **throws `ArgumentOutOfRangeException`** | *(no ranges)* |
| `NoSuchTable[Amount]` | *(no ranges)* | *(no ranges)* |

The throw is the one that could bite a caller: it escapes `IXLDefinedName.Ranges`, a property getter, and it is reachable from ordinary data — load a workbook whose table column has since been renamed. `IsValid` gives no warning, since `ContainsRefError` only tracks a literal `#REF!` in the formula text.

`Sales[Amount]` — the overwhelmingly common form — resolves identically, so the change is narrower than the table makes it look.

## Design

- **Stored as written, resolved on demand.** A table can be resized, renamed or deleted after the formula was parsed, so the area is only true at the moment `Ranges` is asked. Kept in a `readonly record struct` rather than as `StructuredReferenceNode` so that value equality still collapses the same reference written twice in one formula, which the old tuple did.
- **Unresolvable means no range.** Matches what an unknown table already did and what `DependenciesVisitor` does for the same case. Nothing throws. `IsValid` is deliberately left alone: it feeds `ValidNamedRanges()`, which the IO writers and `XLibur.Report`'s `RangeBinder` both filter on, so making a broken reference invalidate the name would quietly change which names survive a round trip. Worth considering separately, not here.
- **`DefinedNameScope`** supplies the resolver's `IStructuredReferenceScope`. A defined name has no anchoring cell, so the caller's existing `anchor` stands in as `FormulaPoint`; that only matters to `[@Column]`, which lands outside the data against the default anchor and so resolves to nothing — the right answer for a name Excel shows as `#VALUE!`. `Worksheet` throws as unreachable: only a table-less reference reads it, and `CollectRefsFactory` does not override the table-less or external-workbook overloads, so one is never collected.
- **Dead code removed:** the `_formula` field and the `Symbol` element of the old tuple were written but never read — `GetExternalRanges` discarded `Symbol` with `_`.

## Tests

`XLibur.Tests/Excel/NamedRanges/DefinedNameStructuredReferenceTests.cs`, 19 cases covering the table above plus `[@Column]`, several references in one formula, a totals reference against a table with no totals row, and that the area follows the table when it is resized.

There was no existing coverage of `IXLDefinedName.Ranges` for a structured reference at all, which is how the wrong answers went unnoticed.

## Pre-existing, not addressed here

`Sales[[#Headers],[#Data]]` — the two-specifier form Excel writes — throws `IndexOutOfRangeException` from inside `ClosedXML.Parser` while the formula is being parsed, before any of this code is reached. I verified it reproduces on unmodified `main`, so it is upstream and out of scope. `StructuredReferenceResolver` already handles the `Headers | Data` combination, so the resolution side is ready if the parser is fixed. Noted in the test class remarks. Happy to raise it upstream against ClosedXML.Parser if wanted.

## Verified

`11700` core tests (4 skipped) and `472` report tests pass; solution builds clean under `TreatWarningsAsErrors`.

### DIFF
--- a/XLibur.Tests/Excel/NamedRanges/DefinedNameStructuredReferenceTests.cs
+++ b/XLibur.Tests/Excel/NamedRanges/DefinedNameStructuredReferenceTests.cs
@@ -1,0 +1,163 @@
+using System.Linq;
+using System.Threading.Tasks;
+using XLibur.Excel;
+
+namespace XLibur.Tests.Excel.NamedRanges;
+
+/// <summary>
+/// What <see cref="IXLDefinedName.Ranges"/> answers for a name whose formula is a structured
+/// reference (<c>=Sales[Amount]</c>).
+/// </summary>
+/// <remarks>
+/// <para>
+/// The area is resolved when it is asked for, through the same
+/// <c>StructuredReferenceResolver</c> that evaluation and the dependency tree use, so a name
+/// honours the <c>#Headers</c> / <c>#Totals</c> / <c>#All</c> specifiers and a column span rather
+/// than always answering with the first column's data.
+/// </para>
+/// <para>
+/// The two-specifier form Excel writes as <c>Sales[[#Headers],[#Data]]</c> is not covered: it
+/// throws inside ClosedXML.Parser while the formula is being parsed, before any of this is
+/// reached, so it is not something resolution can answer for either way. The resolver does handle
+/// the combination once a formula carrying it parses.
+/// </para>
+/// </remarks>
+public class DefinedNameStructuredReferenceTests
+{
+    /// <summary>
+    /// A table at <c>A1:C4</c> — headers <c>A1:C1</c>, data <c>A2:C3</c>, totals <c>A4:C4</c>.
+    /// </summary>
+    private static XLWorkbook TableBook()
+    {
+        var wb = new XLWorkbook();
+        var ws = wb.AddWorksheet("Data");
+        ws.Cell("A1").Value = "Product";
+        ws.Cell("B1").Value = "Amount";
+        ws.Cell("C1").Value = "Tax";
+        ws.Cell("A2").Value = "Widget";
+        ws.Cell("B2").Value = 10;
+        ws.Cell("C2").Value = 1;
+        ws.Cell("A3").Value = "Gadget";
+        ws.Cell("B3").Value = 20;
+        ws.Cell("C3").Value = 2;
+        ws.Range("A1:C3").CreateTable("Sales").SetShowTotalsRow(true);
+        return wb;
+    }
+
+    private static string[] RangesOf(XLWorkbook wb, string formula) =>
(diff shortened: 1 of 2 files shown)


jafin:
Raised the parser limitation upstream: jafin/ClosedXML.Parser#1.

Digging further, it is broader than the two-specifier form I noted in the description. The rule is that in the `[[…]]` inner-reference form, **a specifier list not followed by a column name** throws `IndexOutOfRangeException` from the tokenizer — so `Sales[[#All]]` and `Sales[[#Headers]]` fail as well, while `Sales[[#Headers],[#Data],[Amount]]` parses fine. The unbracketed `Sales[#All]` is unaffected because it returns early on a different branch.

Cause is an unconditional `SkipComma` in `TokenParser.ParseIntraTableReference` that assumes a column always follows the specifier list; its own `Debug.Assert(input[i] == ',')` is the violated invariant, and it compiles out in Release. Nothing on this side needs to change — `StructuredReferenceResolver` already handles `Headers | Data` and `Totals | Data`, so the areas resolve correctly as soon as the formula parses.

PR #370, opened 2026-08-04
fix(io): save spilled and data-table cells as formula results, not constants

jafin:
## Summary

A workbook containing a dynamic-array formula loses its spill on save: the anchor keeps its value and every cell below it renders `#VALUE!` in Excel.

Found on a real export — a template whose `Category` sheet holds `=UNIQUE(tblCategory[Category])` in `E2`, spilling to `E7`. Load and save it and Excel shows one category and five `#VALUE!`s.

Excel writes a spilled cell as a cached formula result:

```xml
<c r="E3" t="str"><v>General Goods &amp; Services</v></c>
```

XLibur wrote it back as a shared-string constant:

```xml
<c r="E3" s="0" t="s"><v>166</v></c>
```

Excel reads a shared-string or inline-string cell inside a spill footprint as content occupying the range. Patching only those five `t="s"` attributes back to `t="str"`, changing nothing else in the package, is enough to make the file load correctly — so the type attribute is the whole cause.

The loader decided whether text was a formula result purely from the presence of an `<f>` element ([`WorksheetSheetDataReader.cs`](https://github.com/XLibur/XLibur/blob/main/XLibur/Excel/IO/WorksheetSheetDataReader.cs)). A dynamic array keeps its formula only in the anchor, so the spilled cells were taken for ordinary text constants, interned into the shared string table, and written back as constants. The writer had no path that could emit `t="str"` for a formula-less cell at all.

**Data tables had the identical bug**, and their non-master cells were silently being turned into shared strings too.

## Changes

- **`WorksheetSheetDataReader`** — a `t="str"` cell is a formula result and never enters the shared string table, whether or not it carries an `<f>`.
- **`SheetDataWriter`** — collects each sheet's cached-result footprints (dynamic arrays *and* data tables) once, and writes the cells inside them the way a formula cell's cached value is written. A sheet with neither pays nothing. Classic array formulas need no entry: the formula slice holds them across the whole range, so their cells already take the formula path.
- **`XLCalcEngine` / `ValueSlice`** — a freshly spilled value is marked as a formula result at the point of spilling, so a workbook built from scratch saves the same as one loaded from a file.
- **`DataTableFormula-Output.xlsx`** — the golden file for `DataTable_MaintainProperties` had recorded the buggy output. Regenerated; the data-table cells now match the Excel-authored input file byte for byte.

## Related Issues

None filed — found in the wild.

## Testing

- [x] All existing tests pass
- [x] Tested manually (described below)

Two regression tests in `SpillEvaluationTests`, covering the fresh-save path and the load/save round-trip:

- `Spill_TextFootprint_SavedAsFormulaResultsNotSharedStrings`
- `Spill_TextFootprint_SurvivesRoundTrip`

Verified against real Excel over COM rather than by inspection alone:

| File | `Category!E3:E7` in Excel |
| --- | --- |
| Export from current `main` | `#VALUE!` ×5 |
| Same package, only `t="s"` → `t="str"` | correct values |
| Template round-tripped through this branch | correct values |

The round-tripped template now produces column E byte-identical to the Excel-authored original.

Full suite green on both target frameworks: **11,826 passed, 0 failed, 4 skipped** (net8.0 and net10.0).

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved dynamic-array and data-table spilled results when saving and reopening workbooks.
  * Correctly retained spilled text values as formula results instead of ordinary text constants.
  * Ensured spilled results remain intact through repeated save and reload cycles.
  * Classic array formulas remain unaffected.


jafin:
Both correct, and they turned out to share a root — pushed in `ee173dce`.

**The evaluation ordering was the worse of the two.** A dynamic array that has never been evaluated has no footprint at all until the save spills it, so with `EvaluateFormulasBeforeSaving` the spill fired from `WriteFormulaCellDirect` *inside* the write loop. That is too late for the footprints, and also too late for the cell enumerator. Saving an unevaluated `=UNIQUE(A1:A3)` in `C1` produced:

```xml
<c r="C1" t="str" cm="1"><f t="array" ref="C1:C2">_xlfn.UNIQUE(A1:A3)</f><v>alpha</v></c>
<!-- no C2 anywhere in the file -->
```

The anchor claims a `ref` the file has no cells for — the spilled cell landed behind the enumerator and was never written, so this was losing cells rather than mistyping them.

Evaluation is now a pre-pass over the sheet, ahead of both the footprint collection and the enumerator, so the two see the final grid. The formulas are collected before any of them is evaluated, since a spill writes into the value slice and can extend the used range — that must not happen under an open enumerator. An array formula is held by every cell of its range so the same instance is collected repeatedly; a second dirty check makes the repeats free and also skips whatever a fallback to full recalculation already cleaned.

**On tracking formulas rather than areas** — done, for the reason you gave. `IsCachedResultCell` now holds the `XLCellFormula`s and reads `Range` at the point of the test. With the pre-pass in place nothing should move mid-write, so this changes no behaviour today; it means a footprint that does move cannot silently go stale and drop its cells back onto the value-only path.

I left the per-cell evaluation in `WriteFormulaCellDirect` as a fallback, since evaluating one sheet can dirty another and sheets are written in order.

**Test:** `Spill_TextFootprint_EvaluatedDuringSave_IsWrittenAsFormulaResults` — sets the formula, never reads it, saves with `EvaluateFormulasBeforeSaving = true`, and asserts both that `C2` exists and that it is typed as a formula result. Confirmed it fails on the previous commit with `<missing cell C2>` rather than passing vacuously.

Full suite green on both TFMs: 11,827 passed, 0 failed, 4 skipped.

PR #373, opened 2026-08-06
perf(io): add template round-trip benchmarks and cut fixed open/save overhead

jafin:
Ports an external performance harness into `XLibur.Benchmarks`, then acts on what it found.

The harness came from profiling a template-driven export in a consuming application, where roughly 300 ms per request went on opening and re-saving a 124 KB workbook — ~10 sheets, ~20 defined names, ~26 data validations — purely to write a single column of lookup values. That cost scaled with neither the rows written nor the amount of data changed, so it is pure per-request overhead.

## Where the harness landed

It arrived as xUnit; this repo uses TUnit, and a multi-iteration timing loop over 20k-row grids does not belong in a serial test suite. `XLibur.Benchmarks` already hosts exactly this shape of work, so it went there as **two** tools, because they answer different questions:

- **`TemplateRoundTripProfile`** — `profile template [path.xlsx]`. Decomposes the round trip into parse/serialise, lookup refresh, bulk grid write, per-cell vs per-column styling, and the full cycle. Fast enough to iterate against, and it accepts a real `.xlsx` (argument or `XLIBUR_PERF_TEMPLATE`) because the generated fixture is far lighter than a genuine reporting template.
- **`TemplateRoundTripBenchmarks`** — the BenchmarkDotNet counterpart, for claims about time.
- **`profile template loop <open|save|roundtrip>`** — runs one phase and nothing else, so a profiler can be attached to it. The table-producing probes interleave six workloads with forced gen2 collections between passes, which makes a captured trace useless.

## What changed in the library

Four inefficiencies in cost that scales with a workbook's structure rather than its data:

- **`LoadWorkbookTheme`** touched `tp.Theme`, materialising the entire theme part — font scheme, format scheme, gradient fills, effect styles — as an object graph, to read twelve hex strings out of `<a:clrScheme>`. Replaced with a raw `XmlReader`. Depth matching keeps the scan off the `a:clrScheme` nested inside `extraClrSchemeLst`, which sits a level deeper and would otherwise overwrite the real scheme.
- **`XLWorksheetContentManager`** ran `Elements<T>().LastOrDefault()` once per slot — thirty-nine filtered traversals of the same child list, each allocating an iterator, per worksheet per save. Now one pass. `XLSheetViewContentManager` likewise.
- **`XLBaseContentManager`** backed its slots with a dictionary and answered `GetPreviousElementFor` with a `Where`/`DefaultIfEmpty`/`MaxBy` chain, allocating on every call while writers call it once per emitted element. The slots are a small contiguous enum range, so a dense array serves better and the lookup walks backwards without allocating. Also drops a box-per-call from casting the enum through `ValueType`.
- **`ConditionalFormattingWriter`** built a hash set, a cast, a concat, an ordering and a list before it could discover that a worksheet carries no conditional formatting at all. Added an early-out; with both collections empty the priority renumbering it skips is a no-op.

## Measurements

BenchmarkDotNet, 10-sheet template with 20 defined names and 26 validations:

| Benchmark | before | after |
|---|---|---|
| `Open` | 4.758 ms / 1.96 MB | 4.695 ms / 1.89 MB |
| `OpenAndSaveUnchanged` | 9.998 ms / 4.28 MB | 10.108 ms / 4.11 MB |
| `RefreshLookupColumn` | 11.851 ms / 4.94 MB | 11.495 ms / 4.77 MB |

**~4% allocation reduction. Time moved less than the run-to-run noise, so no time claim is made.**

Worth stating plainly: an earlier read of the ported harness suggested ~30% gains. That was measurement artefact — the benchmark machine varies by up to 40% between runs of identical code, which is far wider than the effects being chased. The harness was hardened afterwards (nine passes, a `fastest` column, a warm-up that exercises every structural feature) and its docs now say that time claims belong to BenchmarkDotNet and the probe is for locating cost, not proving it moved.

## A correction to the harness's own notes

The harness concluded that `SaveAs` is single-use per workbook instance and disposes the source stream. Both are wrong, measured directly:

- two `SaveAs` calls to two streams that both stay alive succeed;
- the stream a workbook was loaded from is still readable afterwards;
- the failure reproduces on a brand-new workbook, so it is not about loading.

`SaveAs(Stream)` adopts its destination as the workbook origin, which is what lets a later `Save()` write back there. The next `SaveAs` seeks that origin to copy it forward, and a caller who wrapped the destination in a `using` has already disposed it — hence `ObjectDisposedException` from `MemoryStream.set_Position`. The behaviour is coherent; only the diagnostic is poor. No library change made for it here.

## Not done — the biggest remaining win

Every worksheet part is decompressed and XML-tokenised **twice** per load. `LoadWorksheetElements` runs pass 1 with `OpenXmlPartReader` for structural elements, skipping `<sheetData>` row by row — which still fully tokenises it — and then `LoadSheetDataRaw` opens a *fresh* part stream and rescans from the top to find `<sheetData>`. The skip alone measured ~24% of load time.

Fixing it means a genuine single-pass read: a raw `XmlReader` throughout, materialising non-`sheetData` elements into SDK objects on demand. That is a large, correctness-sensitive refactor of the loader touching every worksheet structural feature, so it is called out here rather than attempted.

## Verification

All 11,831 tests pass (net10.0, Release).

## Summary by CodeRabbit

* **Performance**
  * Improved loading and round-trip handling for large or row-heavy workbooks.
  * Reduced overhead when applying cell, column, and multi-property formatting.
  * Added profiling coverage for worksheet layouts, templates, formatting, and lookup updates.

* **Reliability & Security**
  * Hardened workbook XML processing by restricting unsafe external document features.
  * Improved preservation of worksheet content, themes, formatting, and conditional formatting during load and save.

* **Documentation**
  * Added guidance and benchmarks for template round-trip performance.
  * Updated project task tracking and resolved an obsolete pivot-field validation item.

### DIFF
--- a/XLibur.Benchmarks/CellStylingBenchmarks.cs
+++ b/XLibur.Benchmarks/CellStylingBenchmarks.cs
@@ -1,0 +1,148 @@
+using System;
+using BenchmarkDotNet.Attributes;
+using XLibur.Excel;
+using XLibur.Fonts.SixLabors.V1;
+
+namespace XLibur.Benchmarks;
+
+/// <summary>
+/// The cost of styling cells one at a time, split by call shape, against the same writes unstyled.
+///
+/// Run with:
+/// dotnet run -c Release --framework net10.0 --project XLibur.Benchmarks -- --filter '*CellStyling*'
+/// </summary>
+/// <remarks>
+/// Spec 18 task 2. <c>profile template</c> puts per-cell number formatting at ~2.3x the allocation of
+/// formatting the column once, and the repo's own <c>profile alloc</c> shows the formatted 50K-row
+/// create allocating 183 MB against 25 MB unformatted. Neither says *what* is being allocated, and
+/// the answer decides the fix.
+/// <para>
+/// The variants exist to separate three candidates that a single number cannot:
+/// </para>
+/// <list type="bullet">
+/// <item><see cref="ValueOnly"/> is the floor — cell lookup plus the value-slice write.</item>
+/// <item><see cref="StyleFacadePerCell"/> is the natural shape, and pays for the
+/// <c>XLStyle</c>/<c>XLNumberFormat</c> façade chain on each cell.</item>
+/// <item><see cref="StyleAssignedPerCell"/> writes the same style through a pre-built
+/// <see cref="IXLStyle"/>, so it pays the style-slice write without building any façade. The gap to
+/// <see cref="StyleFacadePerCell"/> is what the façades cost.</item>
+/// <item><see cref="TwoPropertiesPerCell"/> sets two properties on one cell. If the façade cost is
+/// per-cell it barely moves from <see cref="StyleFacadePerCell"/>; if it is per-property it roughly
+/// doubles.</item>
+/// <item><see cref="StyleOnColumn"/> is the fast path callers are told to use, for scale.</item>
+/// </list>
+/// </remarks>
+[MemoryDiagnoser]
+public class CellStylingBenchmarks
+{
+    private const int Rows = 20_000;
+    private const string DateFormat = "mmm/yyyy";
+
+    private XLWorkbook _workbook = null!;
+    private IXLWorksheet _worksheet = null!;
+    private IXLStyle _preBuiltStyle = null!;
+    private DateTime[] _dates = null!;
+
+    [GlobalSetup]
+    public void GlobalSetup()
+    {
+        SixLaborsV1FontBootstrap.Register();
+
+        _dates = new DateTime[Rows];
+        var start = new DateTime(2026, 7, 1, 0, 0, 0, DateTimeKind.Unspecified);
+        for (var i = 0; i < Rows; i++)
+            _dates[i] = start.AddMonths(i % 24);
+    }
+
+    [IterationSetup]
+    public void IterationSetup()
+    {
+        _workbook = new XLWorkbook();
+        _worksheet = _workbook.AddWorksheet("Data");
+
+        // Built once, off the measured path: a style carrying the number format under test.
+        var donor = _workbook.AddWorksheet("Donor").Cell(1, 1);
+        donor.Style.NumberFormat.Format = DateFormat;
+        _preBuiltStyle = donor.Style;
+    }
+
+    [IterationCleanup]
+    public void IterationCleanup() => _workbook.Dispose();
+
+    [Benchmark(Baseline = true)]
+    public void ValueOnly()
+    {
+        for (var r = 0; r < Rows; r++)
+            _worksheet.Cell(r + 1, 1).Value = _dates[r];
+    }
+
+    [Benchmark]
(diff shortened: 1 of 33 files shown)


jafin:
Follow-up: the double-parse was investigated, and the conclusion inverts what the PR description assumed.

**The second stream open and rescan are not the problem — they are free.** Attributing the two passes across sheet shapes:

| shape | total load | `SkipSheetData` | open + rescan | row read |
|---|---|---|---|---|
| 10 sheets x 0 rows | 7.52 ms | 0.07 | 0.38 | 0.14 |
| 10 sheets x 100 rows | 14.58 ms | 3.32 | 0.43 | 3.14 |
| 10 sheets x 1000 rows | 84.36 ms | **31.60** | 0.65 | 38.28 |
| 2 sheets x 5000 x 20 | 248.96 ms | **47.17** | 0.12 | 167.95 |
| 1 sheet x 20000 x 20 | 492.69 ms | **116.97** | 0.08 | 267.83 |

Re-opening the part and rescanning to `<sheetData>` costs 0.05-0.65 ms, because everything ahead of `<sheetData>` is small. All of the waste was in *getting past* `<sheetData>` in pass 1 — nearly as expensive as reading the rows for real.

**Cause:** `SkipSheetData` skipped a row at a time. `OpenXmlPartReader.Skip` on a start element issues one raw `XmlReader.Skip` over the subtree and then calls `GetElementInformation`, which runs the element factory and builds an attribute list. Per row that means a throwaway `Row` object plus attributes — 4.9 MB of pure garbage on a 20,000-row sheet.

**Fix (a00258af):** `ReadNextSibling` performs the same subtree skip once and leaves the reader on the next sibling, which is exactly the pattern the loop already used for `CustomSheetViews`. `SheetData` joins that leading skip loop and `SkipSheetData` is deleted. Net diff is -13 lines.

BenchmarkDotNet, mean:

| Benchmark | before | after |
|---|---|---|
| `LoadRowHeavy` (20,000 x 21) | 460.2 ms / 60.14 MB | 405.6 ms / 55.40 MB |
| `Open` | 4.96 ms / 1.89 MB | 4.67 ms / 1.67 MB |
| `OpenAndSaveUnchanged` | 10.23 ms / 4.11 MB | 10.42 ms / 3.89 MB |
| `RefreshLookupColumn` | 11.64 ms / 4.77 MB | 11.49 ms / 4.55 MB |

Allocation is exact and consistently 5-12% lower. Times carry wide error bars on this machine, so treat them as indicative; the shape-sweep probe put the load-time saving at 10-37%, largest on many-rows-few-columns sheets. Added `LoadRowHeavy` because every other benchmark here uses structurally rich but nearly empty sheets and would not have caught this.

**The single-pass loader rewrite floated in the description is no longer worth doing** — it was aimed at the rescan, which turns out to cost almost nothing. Withdrawing that suggestion.

All 11,831 tests pass.

jafin:
@qodo review

PR #397, opened 2026-08-23
refactor(style): one implementation per style interface — delete the deferred facades (spec 23)

jafin:
Implements spec 23 — *One implementation per style interface: delete the deferred facades* — tasks 1 through 5.

`IXLStyle.Batch` was a **second parallel implementation** of seven style interfaces rather than a flush policy on the one facade. Which implementation a caller got was decided inside `XLStyle.Batch`, so `cell.Style.Border.X = v` and `cell.Style.Batch(s => s.Border.X = v)` ran different code for the same assignment, on the same container, through the same interface.

This makes batching a mode on `XLStyle` and deletes the twin — **713 lines**, seven files.

## The defect this closes

`XLBorder.InsideBorder` is container-aware: a cell's `RangesUsed` is a 1×1 range, which has no interior edges, so `RestoreOutsideBorder` captures and restores all four edges and the net effect is correctly nothing. `XLDeferredBorder.InsideBorder` had no container and could not know that — it was byte-identical to its own `OutsideBorder` setter three lines above.

```csharp
cell.Style.Border.InsideBorder = XLBorderStyleValues.Thick;               // no-op    (correct)
cell.Style.Batch(s => s.Border.InsideBorder = XLBorderStyleValues.Thick); // all four (wrong)
```

Fixed by deletion, not by a patch: the batch path now runs the same `XLBorder`.

## Commits

| Task | Commit |
|---|---|
| 1 | `test(style)`: parity test for every border property — **committed red** |
| 2 | `refactor(style)`: `XLStyle` gains the pending-key batching mode (inert) |
| 3 | `refactor(style)`: facades read their key from the style while it batches |
| 4 | `refactor(style)`: route `Batch` through the mode; delete the seven deferred types |
| 5 | `perf(style)`: pending key as loose components in a pooled holder |
| 5 | `perf(style)`: read the component key once per setter, off the pending holder directly |
| — | `fix(style)`: refresh the cached component facades when a batch flushes |

## Task 1 found less than the spec predicted

**13 of 14 argument sets passed, not 12.** Only `InsideBorder` failed. `XLDeferredBorder.InsideBorderColor` does write all four edge colours, but every edge is styleless and `XLBorderKey.Normalize` resets a styleless edge's colour to black when the key is resolved — so the bug was real but invisible in the resulting key. The parity set is narrower than analysed, not wider, so tasks 2–4 were unaffected.

## Task 5 caught two regressions and fixed both at cause

**Pass 1.** The pending key as an `XLStyleKey?` field on `XLStyle` cost **more** than the object graph it replaced, in two separable ways the new benchmark isolates:

* `XLStyleKey` is a large struct and an `XLStyle` is allocated per cell on the ordinary styling path, so holding one inline grew every style. `DirectPerCell` — code this PR does not otherwise touch — went from 32.5 ms / 29.1 MB to **42.0 ms / 42.1 MB**.
* Every `with` on that key copies the whole struct and re-runs the assigned component's `init` accessor, which normalizes and re-hashes it. A six-property batch paid six copies and six component hashes that it then paid again at flush. `BatchPerCell` went from 29.5 ms to **72.0 ms**.

`XLStyle.PendingKey` now holds loose component fields behind one reference, assembles the `XLStyleKey` exactly once at flush, and is rented from a one-deep per-thread cache instead of allocated per batch.

**Pass 2.** That still left batching at a 0.93 `Batch`/`Direct` ratio against a pre-spec 0.90. The deferred twins exposed their component key as a bare field and their setters just wrote it — no equality guard, no container test, one read — whereas the real facades read `Key` twice per assignment (guard, then the `with`), and each read had come to cost two null tests. So: `XLStyle` exposes the pending holder directly, making a facade's `Key` getter one field read and one null test; and every setter reads `Key` once into a local and reuses it. Both behaviour-preserving.

### Numbers

**Read the ratio, not the milliseconds.** This machine drifts hard between sessions: `ValueOnly` in `CellStylingBenchmarks` — which touches no style code at all — moved 8.9% across two sweeps of identical binaries, and pre-spec `BatchPerCell` measured 29.5 ms in one session and 32.3 ms in another. A cross-session millisecond delta under ~10% is not evidence of anything here. The within-run ratio cancels that drift; the allocation columns are exact.

`BatchStylingBenchmarks`, 50,000 cells × 6 properties, net10.0, medians of three runs, all three variants measured back to back in one thermal window:

| Variant | `BatchPerCell` | `Batch`/`Direct` | Batch allocation |
|---|---|---|---|
| Pre-spec (`1b41cadd`) | 32.33 ms | 0.90 | 32.16 MB |
| Branch, before pass 2 | 32.20 ms | 0.93 | 29.49 MB |
| **Branch, final** | **31.41 ms** | **0.90** | **29.49 MB** |

Batch is back to its pre-spec ratio against direct assignment, allocates 8.3% less than the deferred graph did, and no longer provokes gen1 collections on the batch path.

`DirectPerCell` allocation is 29.48 MB against a pre-spec 29.10 MB — a deterministic **+8 bytes per `XLStyle`**, the one `_pending` reference field. Every `CellStylingBenchmarks` variant that builds a per-cell facade gained exactly +0.77 MB over 100,000 cells and the three that do not are byte-identical, which is what makes that attribution certain rather than inferred. Normalised to `ValueOnly` within each session, every per-cell styling ratio is flat or better than pre-spec.

## Also fixed: a retained facade went stale across a batch

**Pre-existing, not fallout from this PR** — both cases below fail identically at the merge-base. Fixed here because the flush is now the one place that can see it.

A component facade the caller obtained before or during a `Batch` stays reachable afterwards, and outside a batch a facade reads its key from its own cached value. That value was never updated at flush, so a retained facade reported pre-batch values — and a later write through it rebuilt the component key from them, silently dropping everything the batch had set:

```csharp
var font = cell.Style.Font;
cell.Style.Batch(s => s.Font.FontSize = 20);
font.Italic = true;                 // FontSize reverts to 11
```

`XLStyle`'s own component getters resync on every access, which is why only a facade the caller *keeps hold of* is affected, and why no existing test caught it.

The border is **refreshed rather than synced**. `XLBorder.SyncValue` drops the pending edge colours when the incoming key differs — right when the style moved underneath the facade, wrong at a flush, where the facade's own writes are what moved it. Blanket-syncing would have broken this, which the direct path gets right:

```csharp
cell.Style.Batch(s => { s.Border.LeftBorder = Thin; s.Border.TopBorderColor = Red; });
cell.Style.Border.TopBorder = Thin;   // must paint the top edge red
```

The colour is held pending because the top edge is styleless, and the direct path keeps it across exactly this transition via `SetKey`. So `XLBorder` gains a `RefreshValue` that assigns the value and leaves the pendings alone. `A_colour_left_pending_by_a_batch_still_applies_afterwards` pins it against the direct-path equivalent, and was checked to fail when the border is synced instead of refreshed — a gate that can actually fail, not one that merely passes.

Benchmarks unmoved: `Batch`/`Direct` ratio 0.91 median of three, batch allocation 29.49 MB.

## Two deviations from the spec text

Both forced by the code, both documented at their site.

1. **`XLStyle.Key` reads through the pending key**, rather than a separate `CurrentKey` leaving `Key` on `Value.Key`. `RestoreOutsideBorder` reads a cell's key back through `XLStyle.Key` mid-edit, so a compound inside-border operation inside a batch would otherwise capture the pre-batch key and restore over assignments the same batch had already made.
2. **Facades read through the style only while it is batching**, keeping `_value` as the source of truth otherwise, rather than dropping `_value` entirely. `XLFont(IXLFontBase)`, `XLFill(null, d)`, `XLBorder(container, null, d)` and their siblings construct a facade over a key that the empty style they fall back to does not hold, so an unconditional read through the style reports the default key for every standalone facade. This also leaves the non-batch path byte for byte what it was.

## Acceptance criteria

| # | Criterion | Result |
|---|---|---|
| 1 | Seven `XLDeferred*.cs` deleted; 713 lines removed | ✅ exactly 713 |
| 2 | One implementation per style interface | ✅ seven files, one each |
| 3 | All 14 parity argument sets pass | ✅ |
| 4 | No public API change | ✅ `PublicAPI.Unshipped.txt` untouched |
| 5 | `XLAlignmentKeyTests`' two throw-tests still pass | ✅ — the batch one now exercises the real facade's validation |
| 6 | Full suite green on net8.0 and net10.0 | ✅ 23,762 passed, 8 pre-existing skips |
| 7 | Batch styling within 10% of its pre-spec median | ✅ at parity — 0.90 ratio, same as pre-spec |

No existing test assertion was weakened.

## Note for spec 20

Spec 20 (style key struct size) now rebases onto this. Its size probe gains a second reason to run: `XLStyle.PendingKey` holds the six component keys inline, so shrinking them shrinks the batch holder too — and `XLStyleKey`'s size is precisely what made an inline pending key untenable here.


jafin:
@coderabbitai review

PR #402, opened 2026-08-24
refactor(drawingml): shared anchor factory and shape-properties writer (spec 16 tasks 2 and 3)

jafin:
Spec 16 tasks 2 and 3, on top of the change-set harness and golden corpus that landed in #401.

Both are behaviour-preserving extractions of shipped code, gated by the suite and by the golden
fixtures #401 captured from the pre-refactor writer. Neither adds a DrawingML capability. **No test
is added, removed or changed by this PR** — the diff is six production files.

`xdr:sp/spPr` and a chart series' `c:spPr` are the same schema type, `a:CT_ShapeProperties`, and all
three anchor forms are the same regardless of what they hold. Both were implemented once, correctly,
in a place only one caller could reach. Spec 15 (shapes and text boxes) needs both, and a second
implementation is how the two copies drift.

## Task 2 — `DrawingAnchorFactory`

`PictureWriter.AddPictureAnchor` built all three anchor forms as three inline blocks, one per
placement.

```csharp
DrawingAnchorFactory.Create(placement, geometry, content) -> OpenXmlCompositeElement
```

returns `xdr:absoluteAnchor`, `xdr:oneCellAnchor` or `xdr:twoCellAnchor`, converting to EMU on the
way. `DrawingAnchorGeometry` carries the drawing's pixel position and size and its two optional
markers. The content is handed in rather than built, which is what lets a caller holding an
`xdr:sp` use the same factory.

The three blocks turned out to be more alike than the spec assumed: it describes them as differing
in "the `xdr:pic` payload and marker sources", but the payloads are byte-for-byte identical. Only
the anchor element and which parts of the geometry it uses actually vary. The picture is therefore
built once.

**The A1 marker fallbacks moved with it and are documented as contract, not implementation detail**,
because a future caller inherits them silently. A missing *from* marker becomes A1 at offset zero; a
missing *to* marker becomes A1 offset by the drawing's own width and height, which is what stops an
unpositioned `MoveAndSize` drawing collapsing onto its own corner. That path is ordinary use, not
misuse — a picture is created `MoveAndSize` with neither marker set, so a picture nobody moved takes
both fallbacks at once.

`PictureWriter` becomes a caller: **net −129 lines**.

## Task 3 — `ShapePropertiesWriter`

```csharp
SetFill(shapeProperties, XLColor?)
EnsureOutline(shapeProperties) -> A.Outline
SetOutlineWidth(outline, double?)
SetOutlineColor(outline, XLColor?)
```

typed over `OpenXmlCompositeElement`, with `BuildColor`, `MapSchemeColor` and the
`CT_ShapeProperties` sequence predicates private to it. `ChartSeriesFormatXml` goes 107 lines
lighter and keeps its thin mask-driven adapters and all its parent-positioning logic.

Three schema facts the original got right and the extraction preserves: fills are a **choice** group,
so setting a colour over an `a:gradFill` removes the whole group rather than appending beside it; the
type is a **sequence** the SDK will not order for you; and existing elements are **mutated in place,
never rebuilt**, so unmodeled children — an `a:round` join, a `cap`, arrowheads — survive an edit
that never mentions them.

## The four redesign rules

1. **Values, never masks.** This was the real work. `SetOutline` used to be
   `SetOutline(C.ChartShapeProperties, XLChartSeries, XLChartSeriesFormat)` and decided for itself,
   from the mask, whether to write the width and whether to write the colour. The mask now stops at
   the chart boundary: `ChartSeriesFormatXml` keeps a twelve-line adapter that reads
   `XLChartSeriesFormat` and asks for the operations it names. Verified by grep — no
   `XLChartSeriesFormat`, no `XLChart*` type and nothing from `DocumentFormat.OpenXml.Drawing.Charts`
   appears anywhere under `DrawingML/`.
2. **Parent positioning stays with the parent.** Where `c:spPr` sits inside `c:ser` is chart
   knowledge; `NewShapeProperties` and its type list are untouched. The shared layer only ever
   operates *within* an `spPr` it is handed.
3. **Colour building moves with the setters.** `BuildColor` and `MapSchemeColor` came across, so
   consumers get theme-colour writing for free.
4. **Behaviour-preserving, proven twice.** Suite green and unmodified, output byte-identical against
   the goldens.

`DrawingUnits` holds the two unit conversions so each has one definition rather than one on each side
of the new boundary. It is deliberately not called `Emu`: `XLibur.Excel.Coordinates.Emu` already
exists, is a length *value* rather than a conversion, and rounds away from zero where this rounds to
even — so the shorter name would have shadowed a type that means something else, and reusing it would
have silently shifted extents at `.5` boundaries.

## What the goldens proved

Every fixture #401 captured from the pre-refactor writer is **byte-identical** after both
extractions. Nothing was re-baselined.

- **`pictures-three-placements`** and **`pictures-default-markers`** cover all three anchor forms and
  both A1 fallbacks, with non-A1 cells and non-zero offsets so the zero-based off-by-one and both
  offset conversions are exercised with values that would fail loudly if wrong.
- **Nine `patch-*` fixtures** cover exactly the paths task 3 rewrote: a fill replaced, cleared and
  themed; an outline widened, recoloured and added; a gradient replaced through the choice group;
  `c:spPr` built from nothing and positioned; the marker's own shape properties.

**29,154 tests green** on net8.0 and net10.0 across all four test projects. Chart 148, picture 75,
drawing 12 — all green and unmodified. `PublicAPI.Unshipped.txt` untouched; every type added is
`internal`.

## Findings reported but deliberately not acted on

Three things surfaced that are out of scope here and are flagged rather than folded in.

**1. Two items on spec 16's task 3 list did not move, on purpose.**

`AppendBeforeExtensionList` names `C.ExtensionList`, a chart-namespace type. Moving it into
`DrawingML/` would have put a chart type into a shared signature — the same category of mistake as
putting a flags enum there.

`InsertAfterLastOf` could have moved, but `ChartElementOrder.InsertOrdered` already exists. Spec 22
introduced it *after* spec 16 was written, and the sibling chart concept modules already use it; it
is a strictly better generalisation of both helpers, and `SeriesChildOrder` already lists
`ChartShapeProperties`, `Marker`, `Smooth` and `ExtensionList` in schema order. Moving the older
helper into `DrawingML/` would have made three ordering implementations where the tree has two. The
right follow-up is folding `ChartSeriesFormatXml`'s two older helpers onto `InsertOrdered` —
chart-side, spec 22's territory. Not free: it needs a `MarkerChildOrder` table and care at eight call
sites, because the two mechanisms differ in their fallback (`InsertAfterLastOf` inserts at index 0,
`InsertOrdered` appends).

**2. A third copy of the pixels-to-EMU conversion exists** at `XLibur/Excel/Drawings/XLPictures.cs`,
logically identical to the one relocated here. Outside `PictureWriter` and outside task 2's scope, so
left alone; a two-line change gated by the same suite whenever wanted.

**3. One branch of the new layer has no test behind it.** `SetOutlineColor(outline, null)` — clearing
a series' line colour. Nothing in the suite sets `LineColor = null`. It was equally untested before
the extraction, since the identical code sat inside the old `SetOutline`, so no path was added; but
it is now a named operation of a shared layer and should not go unmentioned. No test was added
because task 3's gate is the existing suite *unmodified*.

One deliberate behaviour difference, unobservable: the unsupported-placement
`ArgumentOutOfRangeException` now names `placement` rather than `picture`. `XLPicturePlacement` has
exactly three members, so the branch is unreachable without a cast, and nothing asserts the message.

## Summary by CodeRabbit

* **New Features**
  * Improved placement and sizing of pictures, including free-floating, one-cell, and two-cell anchors.
  * Added more consistent handling of picture dimensions and positions when spreadsheet markers are incomplete.
  * Improved chart series and shape formatting for fills, outlines, colors, and line widths.

* **Improvements**
  * Enhanced accuracy when converting measurements for drawings and pictures, supporting more consistent rendering across worksheets.

### DIFF
--- a/XLibur/Excel/IO/Charts/ChartSeriesFormatXml.cs
+++ b/XLibur/Excel/IO/Charts/ChartSeriesFormatXml.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Linq;
 using DocumentFormat.OpenXml;
+using XLibur.Excel.IO.DrawingML;
 using XLibur.Extensions;
 using A = DocumentFormat.OpenXml.Drawing;
 using C = DocumentFormat.OpenXml.Drawing.Charts;
@@ -26,9 +27,6 @@ namespace XLibur.Excel.IO.Charts;
 /// </remarks>
 internal static class ChartSeriesFormatXml
 {
-    /// <summary>EMU per point, the unit DrawingML uses for line widths.</summary>
-    private const double EmuPerPoint = 12700;
-
     /// <summary>
     /// Seeds the model from one <c>c:ser</c> element.
     /// </summary>
@@ -44,7 +42,7 @@ internal static class ChartSeriesFormatXml
         series.SeedLoadedFormat(
             fillColor: ReadSolidFillColor(shapeProperties),
             lineColor: ReadSolidFillColor(outline),
-            lineWidthPt: outline?.Width?.Value is { } width ? width / EmuPerPoint : null,
+            lineWidthPt: outline?.Width?.Value is { } width ? width / DrawingUnits.EmuPerPoint : null,
             markerStyle: ReadMarkerStyle(marker),
             markerSize: marker?.Elements<C.Size>().FirstOrDefault()?.Val?.Value,
             markerFillColor: ReadSolidFillColor(markerFill),
@@ -213,7 +211,7 @@ internal static class ChartSeriesFormatXml
         var shapeProperties = existing ?? NewShapeProperties(seriesElement);
 
         if ((assigned & XLChartSeriesFormat.Fill) != 0)
-            SetFill(shapeProperties, series.FillColor);
+            ShapePropertiesWriter.SetFill(shapeProperties, series.FillColor);
 
         if ((assigned & (XLChartSeriesFormat.Line | XLChartSeriesFormat.LineWidth)) != 0)
             SetOutline(shapeProperties, series, assigned);
@@ -288,7 +286,7 @@ internal static class ChartSeriesFormatXml
             InsertAfterLastOf(marker, markerShapeProperties, typeof(C.Size), typeof(C.Symbol));
         }
 
-        SetFill(markerShapeProperties, series.MarkerFillColor);
+        ShapePropertiesWriter.SetFill(markerShapeProperties, series.MarkerFillColor);
     }
 
     private static void ApplySmooth(OpenXmlCompositeElement seriesElement, XLChartSeries series)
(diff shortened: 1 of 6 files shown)


jafin:
@coderabbitai review

jafin:
@coderabbitai review

PR #409, opened 2026-08-26
fix(grid): row outline levels, GetMaxRowOutline crash and XLColumn.CellCount, on one grid axis

jafin:
Implements [spec 26](docs/specs/26-grid-axis.md). Three shipped defects, each with the test that
would have caught it, then the collapse that removes the duplication which produced them.

## The three defects

**Row outline levels were counted into the column counter.** `XLRow.OutlineLevel` was a verbatim
copy of `XLColumn.OutlineLevel` and kept the column calls, so `Worksheet.IncrementRowOutline` and
`DecrementRowOutline` had **zero callers repo-wide**. For a user's file this meant two things at
once: `sheetFormatPr/@outlineLevelRow` was never written at all, so a sheet with grouped rows
declared a maximum row outline of zero — and every row group instead raised `@outlineLevelCol`, so
a sheet with no grouped columns could still claim column outlines. Because the load path also sets
a row's outline level (`WorksheetSheetDataReader`), this was a **round-trip corruption, not just a
create-path one**: opening a file with row groups and saving it inflated that file's
`@outlineLevelCol`. Files XLibur saved before this carry the wrong attribute; opening and re-saving
them corrects it. Per-row `row/@outlineLevel` was always written correctly and is unaffected — only
the sheet-level summary attributes were wrong.

**`GetMaxRowOutline` threw once that was fixed.** It guarded the dictionary's size rather than the
filtered sequence's, and `Decrement` leaves a zero count behind — so a sheet whose rows were all
ungrouped held a non-empty dictionary of zeroes and `.Max()` ran on an empty sequence. Unreachable
while row outlines were never counted, and reachable the moment the first fix landed, on a sequence
an existing test (`RowTests.UngroupFromAll`) already performs. `WorksheetPartWriter` calls
`GetMaxRowOutline()` on **every** save, so this had to ship in the same commit or the fix would
have shipped a crash. It did.

**`IXLColumn.CellCount()` always returned 1.** Also a verbatim copy of the row half, down to
measuring the *column* span of its address. A column's address runs from row 1 to the last row of a
single column, so both ends carried the same column number and the count was always `c - c + 1`. It
now returns 1,048,576 — one cell per row in the sheet, which is what the row half has always
reported for its own axis. The signature is unchanged; only the value moves. No test asserted it
(the suite's only `CellCount` assertion is on `IXLRangeRow`), which is why it survived.

Both behaviour changes are in `CHANGELOG.md` under `## Unreleased`.

## The collapse

Every row-wise algorithm in the grid had a column-wise twin written out longhand — about 444 lines
existing only because the same algorithm was written twice across five files, and it had drifted
three times in shipped code. There is now one implementation and two thin adapters.

`IGridAxis` is passed as a **generic type argument, never embedded and never held behind an
interface reference**, so the JIT specialises each body per axis. `RowAxis` and `ColumnAxis` are
zero-field `readonly struct`s. This is the constraint spec 21 paid to learn: `grep -rn ': IGridAxis\|IGridAxis [a-z]'`
matches only the two struct declarations.

| File | before | after |
|---|---:|---:|
| `XLRangeInsertHelper.cs` | 226 | 139 |
| `XLRangeShiftHelper.cs` | 144 | 93 |
| `XLWorksheetRangeShifter.cs` | 320 | 222 |
| `XLRangeBase.cs` insert block | — | −41 |
| `XLWorksheet.cs` shift block | — | −47 |

`XLWorksheetRangeShifter`'s page-break and sparkline passes ran in **opposite orders on the two
axes**, with nothing in the file saying whether it mattered. It does not — they touch disjoint
state — but that was an inspection, so it is now a test: the pinning case was written and **run
green against the two-order code before the collapse**, which makes adopting a single order a
proven no-op rather than an assumption. The order is documented on `Shift<TAxis>`, along with the
separate fact that the first three steps are *not* interchangeable.

Out of scope by construction and untouched byte for byte:
`XLibur/Excel/Cells/XLCellFormulaShifter.Legacy.cs` (spec 25's file, 9 further mirror pairs) and
`XLibur/Excel/Cells/ISheetListener.cs` (spec 33's interface — only the *choice* between its four
members is routed through the axis). `git diff --numstat` over `XLibur/Excel/Cells/` is empty.

**No public API change.** `PublicAPI.Shipped.txt` and `PublicAPI.Unshipped.txt` are untouched;
interface widths still read 66 / 65 / 47 / 44.

## Allocation: the insert path got cheaper

The first measurement **regressed +20–33%** on four probes. Cause: `XLAddress` is a `struct` and
`XLRangeAddress` exposes `FirstAddress`/`LastAddress` as that struct, so axis methods typed
`IXLAddress` **boxed on every projection** — once per repository entry per insert. `in XLAddress`
overloads removed it. Worth noting for the next spec: the criterion designed to catch exactly this
looks for an `IGridAxis`-typed *receiver*, and the boxing was on the *argument*.

With that fixed, every probe sits at or below the merge base:

| `profile structural` probe | base | branch | Δ |
|---|---|---|---|
| 0 ranges, 0 formulas | 717–789 MB | 283–401 MB | −50% |
| 1,000 ranges below | 1167 MB | 717–758 MB | −38% |
| 1,000 ranges above (no-ops) | 659–745 MB | 315–405 MB | −45% |
| 1,000 formulas below | 1216.6 MB *(stable)* | 1211.3 MB *(stable)* | −0.4% |
| 1,000 formulas above | 474–574 MB | 224–370 MB | −45% |
| both below | 2023 MB | 1643–1713 MB | −17% |
| 1 batch insert of 1,000 | 3.2 MB *(exact)* | 2.8 MB *(exact)* | −12% |

The improvement is mechanical, not luck: the two longhand copies each read
`thisRangeAddress.FirstAddress`/`.LastAddress` through `IXLRangeAddress` **15 times per call**,
boxing each time; the single implementation hoists them into two locals and reads them **twice**.
Writing the algorithm once removed 13 boxes per shifted range that neither copy could see.

Caveat, now filed as D11 in the conductor's `DEFECTS.md`: `StructuralEditProfile` claims "Bytes are
exact" and **five of its seven probes are bimodal**, swinging up to ±30% run-to-run on identical
code. These figures are modal values over four interleaved runs per side. A single run each way
would have shown a spurious 61% regression on one probe and a spurious 11% improvement on another,
in the same pair of runs.

## The symmetry gate did not bite as the spec specified it

This is probably the most useful thing here for whoever writes the next spec.

Spec 26 required the standing gate to be proven to bite by deliberately mutating each axis. **The
spec's own first mutation passed** — the gate never noticed `ShiftRowHeights` being stubbed out.

The reason: `XLRow.InsertRowsAbove` and `XLColumn.InsertColumnsBefore` move line sizes *themselves*
via `RowsCollection.ShiftRowsDown` / `ColumnsCollection.ShiftColumnsRight`, then call the helper
with `onlyUsedCells: true` — which is the first thing `ShiftRowHeights`/`ShiftColumnWidths` bail
on. Those two methods are **unreachable from the `IXLRow`/`IXLColumn` entry point** that every
originally-specified case used. Confirmed independently: the whole suite under that mutation failed
exactly one test, `RowTests.InsertingRowsAbove4`, which inserts through `ws.Range("3:3")` — the
range form, `onlyUsedCells: false`.

Fix: added `Line_sizes_are_carried_along_identically_on_both_axes`, which seeds row heights and
column widths and inserts through an **entire-line range**. Both mutations then failed and were
reverted. The lesson generalises — a gate written against the convenient entry point can miss a
whole branch of the code it is supposed to guard, and "the test passed under mutation" is the only
thing that reveals it.

No assertion was weakened. One spec-sketched API name did not exist:
`AddConditionalFormat().WhenNotEmpty()` is really `WhenNotBlank()`.

## Acceptance criteria not met

Three, none of them silent:

**Criterion 7 — `XLWorksheetRangeShifter.cs` is 222 lines, target ≤215.** Its *code* is 150 lines,
down from 248; the file is 320 → 222. The 7-line overage is essentially the ordering rationale
described above, which the spec required be documented. Reaching 215 means deleting either that or
the pre-existing ClosedXML#2850 / drop-on-insert rationale the spec explicitly says to keep ("Keep
the full text, once"). I chose the documentation and am flagging the miss rather than trimming
comments to hit a number that exists to prove the file shrank. Happy to cut it if you disagree.

**Criterion 8 — net line reduction 80, target ≥250.** The five algorithm files shrank **324
lines**; `GridAxis.cs` costs **244**. Unreachable by arithmetic: the axis needs **32 members**, not
the 15–20 the spec estimated, because five files each contribute their own axis-dependent
operations. Two full struct implementations of a 32-member interface cannot fit in the ~74 lines
the criterion implies.

**Criterion 10 — pair-count thresholds for the two big files.** Measured with the spec's own
detector:

| File | before | after | target |
|---|---:|---:|---|
| `XLRangeInsertHelper.cs` | 2 | **0** | 0 ✅ |
| `XLRangeShiftHelper.cs` | 1 | 1 | 0 |
| `XLWorksheetRangeShifter.cs` | 6 | **1** | 0 |
| `XLRangeBase.cs` | 7 | 7 | ≤2 |
| `XLWorksheet.cs` | 18 | 17 | ≤2 |

The single pair left in each helper file is the **public entry-point pair the spec told me to
keep**, now one line each delegating to the generic. `XLRangeBase`'s seven and `XLWorksheet`'s
seventeen are all interface obligations — `Row`/`Column`, `RowCount`/`ColumnCount`,
`FirstRowUsed`/`FirstColumnUsed`, the outline pass-throughs — not duplicated algorithm. The
detector counts every Row/Column-named member, so it cannot separate API surface from duplication;
the ≤2 thresholds were calibrated against the wrong quantity.

Also worth knowing: **criterion 14 is unsatisfiable as written.** It asks for the suite green on
net8.0, net9.0 *and* net10.0; `XLibur.Tests` targets `net8.0;net10.0` only. The library still
builds clean on all three.

## Regenerated binary fixture

`XLibur.Tests/Resource/Examples/Misc/Outline.xlsx` is regenerated, and it is the one binary in this
diff. **The old fixture encoded the defect.** `XLibur.Examples/Misc/Outline.cs` groups rows to
outline level 2, and the committed reference workbook has no `@outlineLevelRow` attribute at all —
so `MiscTests.Outline` was asserting that XLibur does not emit the attribute Excel emits. Once row
outlines reached the row counter, that test failed.

I extracted both packages and diffed every part before replacing it. The only substantive
difference is `outlineLevelRow="2"` being added to `sheetFormatPr`; the other two differing parts
are `_rels/.rels` and the core-properties `.psmdcp`, which carry random per-save GUIDs that
`ExcelDocsComparer` ignores — it reported only `/xl/worksheets/sheet1.xml :NonEqual`.

## Testing

`dotnet test XLibur.Tests/XLibur.Tests.csproj` (unfiltered, so both TFMs run):

```
net8.0  passed
net10.0 passed
total: 28264  failed: 0  succeeded: 28254  skipped: 10
```

New tests: `OutlineRoundTripTests` (4 cases — fresh save, reload-and-resave, the column control,
and the ungroup-everything crash), `GridAxisSymmetryTests` (13 cases — a transpose gate across
insert, delete, live-range repositioning, conditional formats, data validations, line sizes, and
page-break/sparkline survival), plus one case in `ColumnTests`.

## Notes for review

- Tasks 1–3 are three independent commits, each carrying the test that would have caught its
  defect, so they stand on their own if the collapse is contentious.
- Tasks 4–8 are in ascending order of risk, `XLWorksheetRangeShifter` last.
- Nine commits, one per task, plus the docs commit. Each is individually green.
- `docs/specs/` is deliberately only partly synced: the four files spec 26 owns or touched. Spec 28
  is in flight and its spec text should land with its own PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed row grouping and ungrouping metadata updates to prevent incorrect column outline data and save-time errors.
  * Corrected column `CellCount()` to report the full worksheet row capacity of 1,048,576 cells.
  * Improved consistency when inserting, deleting, and shifting rows or columns, preserving related formatting and worksheet features.

* **Documentation**
  * Updated specifications and project roadmap statuses to reflect completed grid-axis work.


jafin:
Thanks — all four addressed in 7cd215e7.

**`XLRow.CopyRows` outline level (the merge-risk item) — valid, fixed, and it was worse than described.** `InsertRowsBelow` copies the source row's properties onto the new one and wrote `_outlineLevel` straight to the backing field, so the worksheet's outline counter never saw it. Verified the consequence rather than inferring it: group rows 1–2, insert below row 2, ungroup 1–2, save — row 3 reports `OutlineLevel = 1` while `sheetFormatPr` declares no `@outlineLevelRow` at all. Excel gets a sheet whose summary contradicts its rows.

The part worth flagging: **`XLColumn.CopyColumns` has the identical line and reproduces identically**, and the column counter has been live all along — so this is a pre-existing shipped defect on the column axis, not just a consequence of this PR. Task 1 merely gave the row axis the same exposure. Both are now assigned through the property, which also correctly decrements any level the target line already held. `OutlineRoundTripTests.An_inserted_copy_of_a_grouped_line_is_counted_on_both_axes` covers both axes and fails before the change on each.

That the same one-line defect sat unnoticed in both halves of a copy-pasted pair is, unhelpfully, this PR's entire thesis.

**Spec 33 "halved both shared files" — valid, reworded.** The figures I quoted contradicted the claim: 320 → 222 is a third, and `XLWorksheet.cs` lost 47 lines of ~1,957, about 2%. I had carried over spec 26's own prediction that it would halve them. It now states the real reduction and notes the prediction was wrong.

**Results section / criterion 14 — valid, added.** It now names net8.0 and net10.0 explicitly and records that criterion 14 cannot be met: it asks for net9.0, which `XLibur.Tests` does not target, though the library builds clean on all three.

**Duplicate 26.10 tasklist row — valid, removed.** Stale unchecked copy of the completed entry.

Full suite after the change: **28,266 tests, 0 failed**, both TFMs.

On docstring coverage at 28%: the uncovered members are the `RowAxis`/`ColumnAxis` adapter bodies, which are one-expression implementations of an interface whose members are documented once on `IGridAxis`. Documenting each twice would restore exactly the duplication this PR removes, so I have left them.

Issue #313, opened 2026-08-01
ClosedXML.Parser dependency: we have no route to ship a parser fix, and six workarounds already in the tree


jafin:
Recording findings for a later decision. No action proposed yet — this is a place to collect what is currently scattered across six code comments and two upstream repositories.

Surfaced while fixing #274-adjacent work in #311, where a defined name holding `Sales[[#Headers],[#Data]]` throws `IndexOutOfRangeException` out of the parser. The parse bug itself is diagnosed and fixed; the point of this issue is that **we have no way to ship that fix**.

## The gap

XLibur references `ClosedXML.Parser` **2.0.0** from NuGet, published by the ClosedXML org. The whole formula layer sits on it — calc engine, defined names, formula shifting, dependency tree, `XLCellFormulaShifter`, `FormulaReferences`, `StructuredReferenceResolver`.

A fix on our fork does not reach XLibur. Only a new package version does, and there is currently no route to one:

| | |
|---|---|
| Last release | **2.0.0** — 2025-04-03 (~16 months ago) |
| Last upstream commit | 2025-10-22 (~9 months ago) |
| `develop` vs `2.0.0` tag | **11 commits ahead**, all of them a ground-up rewrite: *"Add a hand-made lexer for 3.0"*, *"Add a basic Pratt parser"*, *"Pratt parser - parse ident"*, *"IdentParselet can recognize 3D references"* |
| 2.0.x maintenance branch | none |
| Open upstream issues | 3, two filed in the last two months by XLibur-adjacent users |

So `develop` is a half-finished 3.0 and there is no line a 2.0.1 could come from. A PR upstream today lands in the rewrite, not in anything shippable.

Useful detail for whoever picks this up: `src/ClosedXML.Parser/TokenParser.cs` is **byte-identical between the `2.0.0` tag and `develop`**, so a patch applies equally to either base.

## Related issues, ours and upstream

- ClosedXML/ClosedXML.Parser#31 — the `IndexOutOfRangeException` above (raised by us, 2026-08-01)
- ClosedXML/ClosedXML.Parser#30 — `ToR1C1` emits `NegativeSign` U+2212 under `sv-SE`/`fi-FI`/`nb-NO` and `ToA1` cannot parse it back, so the R1C1 round trip fails (raised by @joakimriedel, 2026-06-01). Adjacent to our own culture work in #305.
[...]

PR #474, opened 2026-09-13
fix: the formula parser comes from our fork, now at 3.0.0

jafin:
## Summary

XLibur referenced `ClosedXML.Parser` 2.0.0, which has no maintained 2.0.x line, so there was no way to ship a parser fix (#313). This PR moves the reference to `XLibur.ClosedXML.Parser`, our fork, and takes it at 3.0.0. The assembly name and the `ClosedXML.Parser` namespace are unchanged, and so is XLibur's public API.

### Switch to the fork (written against 2.1.0-alpha.276)

- The `PackageReference` points at `XLibur.ClosedXML.Parser`.
- Pins three fork fixes that are reachable from XLibur's API, each confirmed to fail on 2.0.0:
  - A keyword-list structured reference in a defined name threw `IndexOutOfRangeException` on load.
  - An R1C1 round trip failed under a culture whose negative sign is U+2212.
  - Sheet names were quoted the way the formula bar shows them, not the way the file format needs.

### Move to 2.1.0-beta.293

beta.293 changes the parser API, so this commit adapts XLibur to it:

- `RefModVisitor` is now `FormulaModifier`, for the two rename visitors and `SafeModifyA1`.
- `IAstFactory` gains `SheetErrorNode`, `ExternalDynamicDataExchange` and `DynamicDataExchange`. `CollectVisitor` has them as no-ops. The evaluator gives `Sheet1!#REF!` its error and a DDE reference a `NotSupportedNode`.
- `Sheet1!#REF!` no longer goes through `ErrorNode`, so `FormulaReferences` flags it in `SheetErrorNode` too. Without that, a defined name holding it counts as valid. With the line disabled, two `NamedRangeDeletionTests` fail.
- The parser now lexes `#CALC!`, `#FIELD!` and the other newer error values. `XLError` has no member for most of them, so `GetErrorValue` threw `InvalidOperationException` out of `Parse`. It now throws `ExpressionParseException`, the documented type for a formula that cannot be read.

### Move to 3.0.0

3.0.0 adds 11 fix commits to beta.293 and needs no code change in XLibur. The two that reach XLibur users are the last two items below, each pinned by a test.

### What users will see

These are in `CHANGELOG.md` under Unreleased:

- Renaming or deleting a sheet or table rewrites a formula only where it names that sheet or table. The rest keeps its quoting, its one-cell areas and its leading whitespace. This comes from the fork's changelog; XLibur calls the same `FormulaConverter.ModifyA1` path.
- A DDE formula parses. Evaluating one throws `NotImplementedException`.
- `#SPILL!` parses as `XLError.SpillRange`.
- `#GETTING_DATA` throws `ExpressionParseException`, not `InvalidOperationException`.
- Renaming a sheet to a cell-like name such as `PWD1` quotes it as the first sheet of a 3D reference: `'PWD1:Last'!A1`. The bare form doesn't read back.
- `(A1) B2`, a space intersection after braces, parses. Evaluating it throws `NotImplementedException`, as any range intersection does.

## Related Issues

Refs #313

## Testing

  - New: `Constant_error_without_an_XLError_is_a_parse_error`, 4 cases. I watched it fail with `InvalidOperationException` before the fix.
  - New: a `#SPILL!` case in `Constant_can_be_error`.
  - `Reference_can_be_dynamic_data_exchange` is no longer skipped, and it passes.
- [x] All existing tests pass. On net10.0, after rebasing onto `main`: XLibur.Tests 14,493 (4 skipped), XLibur.Report.Tests 481, XLibur.Fonts.SixLabors.Tests 31, XLibur.Fonts.SkiaSharp.Tests 37.
- [ ] Tested manually

Two comments about what the parser can't read were wrong on beta.293. I checked each against the package before correcting it:

- `D3:@A1:C2` now parses, but its dependencies are not pinned in `DependencyTreeTests` yet. The comment there now says so.
- A bang name (`!Total`) now parses, and `XLDefinedName` said it could not. Evaluating one still throws, because `FormulaParser.BangName` answers with `NotSupportedNode`.

## Checklist

- [x] No new warnings introduced (Debug build of `XLibur.slnx`, 0 warnings)

### DIFF
--- a/Architecture.md
+++ b/Architecture.md
@@ -19,7 +19,7 @@ directly.
 | Target frameworks  | net8.0, net9.0, net10.0                                     |
 | License            | MIT                                                         |
 | Core dependency    | DocumentFormat.OpenXml 3.4.1                                |
-| Parser             | XLibur.ClosedXML.Parser 2.1.0-alpha.276                     |
+| Parser             | XLibur.ClosedXML.Parser 2.1.0-beta.293                      |
 | Font handling       | SixLabors.Fonts 1.0.1                                       |
 | Spatial indexing   | RBush.Signed 4.0.0                                          |
 | Number formatting  | ExcelNumberFormat 1.1.0                                     |
--- a/Architecture.md
+++ b/Architecture.md
@@ -19,7 +19,7 @@ directly.
 | Target frameworks  | net8.0, net9.0, net10.0                                     |
 | License            | MIT                                                         |
 | Core dependency    | DocumentFormat.OpenXml 3.4.1                                |
-| Parser             | XLibur.ClosedXML.Parser 2.1.0-beta.293                      |
+| Parser             | XLibur.ClosedXML.Parser 3.0.0                               |
 | Font handling       | SixLabors.Fonts 1.0.1                                       |
 | Spatial indexing   | RBush.Signed 4.0.0                                          |
 | Number formatting  | ExcelNumberFormat 1.1.0                                     |
--- a/Architecture.md
+++ b/Architecture.md
@@ -19,7 +19,7 @@ directly.
 | Target frameworks  | net8.0, net9.0, net10.0                                     |
 | License            | MIT                                                         |
 | Core dependency    | DocumentFormat.OpenXml 3.4.1                                |
-| Parser             | ClosedXML.Parser 2.0.0                                      |
+| Parser             | XLibur.ClosedXML.Parser 2.1.0-alpha.276                     |
 | Font handling       | SixLabors.Fonts 1.0.1                                       |
 | Spatial indexing   | RBush.Signed 4.0.0                                          |
 | Number formatting  | ExcelNumberFormat 1.1.0                                     |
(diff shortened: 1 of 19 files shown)


jafin:
Addressed in d60562be. I checked both outside-diff findings against the current code first.

**`FormulaParserTests.AssertCanParseButNotEvaluate`: fixed.** The helper accepted any `Exception` and never read `notSupportedMessage`. It now requires `NotImplementedException` with that exact message. Both callers pass: the DDE test (now run on both the displayed and the stored form) and the range intersection test.

**`FormulaReferences.CollectRefsFactory`: fixed, with one change to what was suggested.**
- Valid as reported. On 2.1.0-beta.293 a defined name holding `#CALC!` was stored as understood, while the old parser had rejected it.
- `ErrorNode` now calls `XLErrorParser.ParseFormulaError`, which the evaluator's `GetErrorValue` also uses. `ErrorValue` calls it too, because an error inside an array constant (`{1,#FIELD!}`) never reaches `ErrorNode`.
- The change: letting `ExpressionParseException` escape `TryForFormula` would also have escaped the lenient load path, and one such name would again stop a workbook from loading. So `TryForFormula` catches it and hands it back as the failure. `RejectionOf` returns it unwrapped. The setter throws, and a load keeps the name verbatim.
- `SheetErrorNode` is not changed. The parser only ever gives it `#REF!`, so a check there could never fail.
- New test cases:
  - `#CALC!` and `{1,#FIELD!}` are refused by the setter. Both failed before the fix.
  - A workbook with a `#CALC!` name still loads, with its text kept.

**SonarCloud new-code coverage.** Four new lines had no test. Each now has one:
- `FormulaParser.SheetErrorNode`: `Sheet1!#REF!` evaluates to `#REF!`.
- `FormulaParser.ExternalDynamicDataExchange`: the stored form `[1]!'…'` parses.
- The two DDE members of `CollectVisitor`: a defined name holds both DDE forms.

All four test projects pass on net10.0: 14,491 (4 skipped), 481, 31 and 37. The build has 0 warnings.

PR #511, opened 2026-09-14
fix: a pivot table keeps its link to its conditional formats through a save (#507)

### LINKED ISSUE #507: A pivot table loses its link to its conditional formats on every save (D81)
## What happens

A pivot table loses its link to its conditional formats on every save, whether or not anything was edited.

`PivotTableDefinitionPartWriter2` rebuilds the pivot table's `x14` extension from the model and leaves out `x14:conditionalFormats`. That element is how the pivot table refers to its conditional-format rules, by rule id. The rules themselves stay in the worksheet's `x14` extension, marked `pivot="1"`.

## Reproduce

Use `XLibur.Tests/Resource/Other/SheetLifecycle/chartex-pivotcf-before.xlsx`, an Excel-authored file added by PR #503. On sheet `Other` it has a pivot table with a conditional format.

1. Load the workbook.
2. Save it, with no edit.
3. Look at the saved pivot table definition: its `x14:conditionalFormats` is gone.

## Expected

A load and save without edits keeps the pivot table's `x14:conditionalFormats`.

## Notes

This is defect **D81**, found while fixing #497/#498 (PR #503). It's a round-trip defect, not a sheet-lifecycle one, so PR #503 leaves it out of its comparison.

jafin:
## Summary

A pivot table lost its link to its conditional formats on every save, with or without an edit (defect **D81**). The link is lost; the rules themselves stay on the sheet. `PivotTableDefinitionPartWriter2` rebuilds the pivot table's `x14:pivotTableDefinition` from the model, and nothing loaded `x14:conditionalFormats`. That list is how the pivot table names its rules, by id; the rules live in the worksheet's `x14` extension with `pivot="1"`.

Closes #507

## Fix

The list is now **modelled**, not carried as raw XML. The pivot table's main-schema `<conditionalFormats>` is already modelled with the same fields, pivot areas already have a reader and a writer, and the other x14 pivot data (alt text, `hideValuesRow`) is modelled too.

- `XLPivotExtensionConditionalFormat` (new): one entry, holding the rule id, priority, scope, type and pivot areas.
- `XLPivotTable`: holds the list, next to the main-schema list.
- `PivotTableDefinitionPartReader`: loads the list where it already reads the other x14 attributes. An entry with no id or no priority is skipped; the load does not fail.
- `PivotTableDefinitionPartWriter2`: writes the list back. The code that writes scope, type and areas is now shared with the main-schema list, whose output does not change.

The id and priority are written as they were loaded. The worksheet writer keeps a `pivot="1"` rule as loaded, with the same id and priority, and it writes the sheet before its pivot tables. So after a save, each side still names the other. Excel's `chartex-pivotcf-rename-after.xlsx` and `-delete-after.xlsx` show the pivot table's list unchanged by the edit.

The main-schema `<conditionalFormats>` was not lost. Its priorities stay consistent, because the sheet writer renumbers them before the pivot writer runs. A new test pins this.

## Tests

- `PivotConditionalFormatLinkTests` (new, 10 tests). All 9 that test the fix failed before it; the 10th pins behaviour that was already correct.
  - A load and a save keep the list exactly as Excel wrote it, and a reloaded file saves it again.
  - Every rule id in the list is a `pivot="1"` rule on the sheet, with the same priority. This holds with no edit, after a rename or a delete of `Data`, and after a row or a column insert on `Data` or on `Other`.
  - The main-schema list uses the renumbered priority.
- `SheetLifecycleChartExFixtureTests` now also compares the pivot table's list with Excel's rename-after and delete-after files, including the cases with two saves.

| Project | net8.0 | net10.0 |
|---|---|---|
| XLibur.Tests | 15,435 passed, 4 skipped | 15,435 passed, 4 skipped |
| XLibur.Report.Tests | 508 passed | 508 passed |
| XLibur.Fonts.SixLabors.Tests | 31 passed | 31 passed |
| XLibur.Fonts.SkiaSharp.Tests | 37 passed | 37 passed |

After the rebase onto `6acc413a` (#510), `PivotConditionalFormatLinkTests` was run again: 10 of 10 passed.

`/code-review medium`: no findings.

## Not covered, and why

- **Inserts: no Excel fixture.** The insert tests assert only that the link survives. The list names pivot fields and items, not cells, so an insert has nothing in it to move.
- **The range of a pivot rule does not move after a structural edit.** On the sheet side, #509 keeps a pivot rule's `xm:sqref` as loaded. After a row inserted above the pivot table on `Other`, the rule stays on `G2:G4`. A fixture of this workbook with such a row inserted would show what Excel does.
- **Data-bar pivot rules are not verified.** No fixture shows whether Excel also writes an entry here for a data-bar rule.
- **Left as they were:** a copied pivot table carries neither list, because the rules they name belong to the source sheet. The entry's own optional `extLst` is not carried; no file here has one. The `xpdl:pivotTableDefinition16` extension is still dropped on save; that is a separate loss.
- **Gaps found in review, from earlier commits, not changed here:**
  - A rule kept only in `x14` keeps the priority it was loaded with, while the sheet renumbers its modelled rules from 1. Two rules can end up with the same priority. This does not break the pivot link, which is by id.
  - Deleting a pivot table leaves its `pivot="1"` rule on the sheet.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pivot tables now preserve links to their conditional formatting rules when workbooks are saved and reopened.
  * Conditional formatting links remain accurate after related sheet renames, deletions, and row or column insertions.
  * Conditional formatting rule priorities and identifiers are retained across repeated saves.


jafin:
Fixed in `de3e87d7`.

**The finding is valid.** `LoadExtensionList` looked for `x14:pivotTableDefinition` only in the first `<ext>` of the pivot table's `extLst`. An `extLst` holds its extensions in any order, so another extension ahead of the x14 one hid the `x14:conditionalFormats` list, and a save cut the link this PR adds. The lookup was already on `main`. In the same case it also missed the alt text, the description, `enableEdit` and `hideValuesRow`.

**Fix.** The reader now takes the extension that holds `x14:pivotTableDefinition`, wherever it sits in the list (`PivotTableDefinitionPartReader.LoadExtensionList`). The writer has no such lookup, because it rebuilds the x14 extension from the model.

**Test.** I added `PivotConditionalFormatLinkTests.The_list_is_kept_when_another_extension_comes_before_it`.
- The fixture is derived from the Excel-authored `chartex-pivotcf-before.xlsx`. In that file Excel wrote the x14 extension first and its own `xpdl:pivotTableDefinition16` extension second. The test moves the x14 extension to the end, so that a real Excel extension comes before it. It asserts this precondition first.
- It then saves and checks two things. The pivot table's `x14:conditionalFormats` must match Excel's list in full, and the rule it names must still be the `pivot="1"` rule on the sheet, with the same priority.
- Before the fix it failed with `no x14:conditionalFormats`. It passes now.

**Validation.**
- All 100 pivot table tests pass on net10.0.
- The branch is rebased onto `main`, which now includes #512.
- On the rebased branch, the full XLibur.Tests suite passes on net8.0 and net10.0: 30,952 passed, 0 failed, 8 skipped (the same skips as before).

I also added a sentence about this case to the CHANGELOG entry.

PR #569, opened 2026-09-16
fix: a formula keeps a fullwidth colon it really had when a colon in a column name is hidden from the parser (#557)

### LINKED ISSUE #557: A formula loses a real fullwidth colon when the colon protection is on
## What happens

A formula that contains a real fullwidth colon, U+FF1A, loses it when another part of the same formula needs the colon protection.

`FormulaText` protects a colon inside a single-bracket table column name by replacing it with `ColonPlaceholder`, which is the character `：` (U+FF1A), so the parser does not read it as the range operator (`FormulaText.cs:99`, `:358`, `:443`). Afterwards it restores the placeholder over the whole text:

```csharp
rewritten = wasProtected ? result.Replace(ColonPlaceholder, ':') : result;   // :180
converted = wasProtected ? result.Replace(ColonPlaceholder, ':') : result;   // :211
```

The restore cannot tell a placeholder from a U+FF1A the formula really had. So every U+FF1A in the text becomes an ordinary colon, wherever it sits: in a sheet name, in a string, or in a defined name.

`Table1[a:b]&'x：y'!RC` converts to `...x:y!C3`. The sheet name is no longer quoted and is not valid, because the quotes were written for `x：y` and the name is now `x:y`.

## Reproduce (to confirm)

1. Convert a formula that has both a colon inside a table column name and a U+FF1A in a quoted sheet name or a string, in either direction (`FormulaText.TryConvert`).
2. Expected: the U+FF1A is still there in the result, and any sheet name is still quoted correctly.

Found while working on #542 (PR #549): an independent reviewer of the new `A1Template` ran this input. `TryConvert` behaved this way before #549, and the template reproduces it exactly, so it is not a regression.

## Expected

The protection only restores the colons it replaced, and leaves a U+FF1A the formula really had. For example, protect with a character that cannot appear in a formula, or record where each replacement was made and restore only those positions.

## Notes

- Only a formula that has a colon inside a single-bracket column name turns the protection on, so the rest of the time U+FF1A is safe.
- `A1Template` applies the same restore (`A1Template.cs:325`), so both paths are fixed together.

jafin:
## The bug

`FormulaText` hides a colon inside a single-bracket table column name from the parser, which would otherwise read it as the range operator, by replacing it with `ColonPlaceholder` — which **is** the character `：` (U+FF1A). It then restored over the whole text:

```csharp
rewritten = wasProtected ? result.Replace(ColonPlaceholder, ':') : result;   // :180
converted = wasProtected ? result.Replace(ColonPlaceholder, ':') : result;   // :211
```

The restore could not tell its own placeholder from a U+FF1A the formula really had, so every U+FF1A in the text became an ordinary colon — in a quoted sheet name, in a string, in a defined name, or in another column name. `Table1[a:b]&'x：y'!RC` converted to a formula naming the sheet `x:y`, which the workbook does not have.

Only a formula that has a colon inside a single-bracket column name turns the hiding on, so no other formula was ever at risk. This is not a regression from #549; the behaviour predates it.

## The approach, and the one I rejected

**Rejected — "a character that cannot appear in a formula."** That claim cannot be established. A formula can hold any character, in a string literal, in a quoted sheet name or in a column name. I probed the parser with 13 candidates (the fullwidth colon, the small and vertical colon forms, the modifier letter colon, the ratio sign, private-use characters, a noncharacter, and U+FFFD). It accepted every one inside a column name and kept it character for character. So no character is safe by its own nature, and any fixed choice reproduces this bug for some input.

**Rejected as the primary mechanism — recording the positions replaced.** The positions are known in the *input*, but every restore happens on the parser's *output*, whose positions differ: the conversion writes each reference again and changes lengths. Mapping input positions onto output positions would mean tracking every rewrite `ClosedXML.Parser` makes, inside the converter.

**Chosen — pick the placeholder for each formula, as a character that formula does not hold.** This keeps what position tracking was meant to guarantee, that only the colons XLibur replaced are put back, without needing a position map, and it can be shown to be exact:

- What the conversion writes is made only of characters copied from the input and the ASCII a reference is written with (`$ A-Z 0-9 : ! ' [ ] #`).
- So a non-ASCII character absent from the input cannot appear in the output unless the protection put it there.
- Restoring that character is therefore exact.

The placeholder is the fullwidth colon when the formula does not hold one — the usual case, one scan, no change in behaviour and no extra cost — and otherwise the first private-use character (U+E000–U+F8FF) the formula does not hold.

One residual is documented in the code: a formula that holds the fullwidth colon *and* all 6,400 private-use characters, so at least 6,401 characters long, leaves nothing to choose, and falls back to today's behaviour. No same-length placeholder can do better for a text that already uses every character, and the length has to be preserved because every `SymbolRange` the parser reports indexes the original text.

## Both paths are fixed

- **`FormulaText`** — `ProtectStructuredRefColons` now reports the character it chose instead of a `bool`, and `TryConvert`, `TryRewrite`, `TryWalk` (through `ColonRestoringFactory`) and `TryCreateA1Template` all restore that character.
- **`A1Template`** (`:325`) — the restore over the literals is **gone**, not merely given the new character. The literals are built from the caller's own unprotected text, so they hold the real colon already, and that restore could only ever corrupt a character the formula really had. Only a slot's sheet prefix is taken from the text the parser read, so only it is restored.

## How I verified it

New tests cover `TryConvert` in both directions, `TryRewrite` (through `AddFuturePrefixes`), the column names `TryWalk` hands the calc engine, and the `A1Template` path, each with a real U+FF1A in a quoted sheet name, in a string, in a second column name, and beside a hidden colon in the same name (`Table1[a:b：c]`). All 13 failed before the change and pass after it.

One expectation needed correcting while I worked. The conversion writes each reference again, and the parser decides there which sheet names need quotes, so `'x：y'!RC` legitimately converts to `x：y!C3` without them, whether or not anything in the formula turned the protection on. A control test pins that down, with `'My Sheet'` keeping its quotes, and a round-trip test shows the converted text still names the sheet it started with. What #557 changed is the name itself, which used to come back as `x:y`.

There is also a guard test that every character the protection can choose survives the parser inside a column name, checking both ends of the private-use range, because the search walks it from one end to the other.

All four test projects on `net10.0`:

| Project | Result |
| --- | --- |
| XLibur.Tests | 16,052 passed, 0 failed |
| XLibur.Report.Tests | 511 passed, 0 failed |
| XLibur.Fonts.SixLabors.Tests | 31 passed, 0 failed |
| XLibur.Fonts.SkiaSharp.Tests | 37 passed, 0 failed |

Closes #557


jafin:
## Code review round (medium) — three findings, all addressed

**1. Medium — the placeholder guarantee did not cover `TryRewrite`. Correct, and reachable from the public API.**

`PickPlaceholder` guaranteed only that the chosen character was absent from the *input*. That is enough for `TryConvert` and `TryCreateA1Template`, whose output is built from the input's characters plus the ASCII a reference is written with. It is not enough for `TryRewrite`: a `FormulaModifier` writes names into the result that were never in the input. So renaming a sheet to `x：y`, in a workbook whose formula also hides a colon in a column name, reproduced #557 in the rename path — the very invariant I had written down was false there.

I confirmed it rather than accept it on reasoning alone, since reasoning is what produced the wrong invariant in the first place. With the fix backed out, the new test fails:

```
Expected to contain "x：y"
but received "SUM(Table1[Start: Date])+x:y!A1"
```

`FormulaModifier` is a `ClosedXML.Parser` type and cannot be extended, but `RenameRefModVisitor` is ours. It now implements a small `IInjectsText` interface reporting the new sheet and table names, and `TryRewrite` passes `modifier as IInjectsText` into the placeholder choice. No call site changed. The future-function remap writes only ASCII and every candidate placeholder is outside ASCII, so it needs nothing.

**2. Low — the remarks stated the invariant too broadly.** Fixed. They now separate what a conversion writes from what a rewrite writes, and name the injected-text route, so the next change does not inherit the wrong guarantee.

**3. Low — the private-use test never invoked `PickPlaceholder`.** Correct: the formula under test held no ASCII colon, so `ProtectStructuredRefColons` took its early return and the search never ran. The test's doc comment claimed more than the test did. I corrected the claim, and added two tests that do drive the search:

- U+FF1A and U+E000–U+F8FE all taken, so the last private-use character must be chosen — this exercises the end of the range the search walks.
- every candidate taken, pinning the documented fallback and the one residual case.

Not treated as defects, agreeing with the reviewer: the final `return ColonPlaceholder` needs a formula holding U+FF1A and all 6,400 private-use characters, and is the deliberate documented choice; and the small `RestoreColons` duplication in `A1Template.Build`.

Re-verified after the fixes, all four projects on `net10.0`:

| Project | Result |
| --- | --- |
| XLibur.Tests | 16,055 passed, 0 failed |
| XLibur.Report.Tests | 511 passed, 0 failed |
| XLibur.Fonts.SixLabors.Tests | 31 passed, 0 failed |
| XLibur.Fonts.SkiaSharp.Tests | 37 passed, 0 failed |

